PR #17, opened 2015-11-18
Bugfixes in State-Checking, D.C. time-shifting, and AlStatusCode enumerations

AversanJeff:
1.	Function ecx_statecheck will not see an EtherCAT State ‘Error bit ‘
In function ‘ecx_statecheck(ecx_contextt *context, uint16 slave, uint16 reqstate, int timeout)’, there is a line :
state = rval & 0x000f; /* read slave status */
This will get the final 4 bits of an EtherCAT state, but will NOT catch any Error bit, if present.  Consequently, if you want to check a state is (for example) “Safe-Op” but the actual state is “Safe-Op Error”, the function will incorrectly report it is in “Safe-Op”.  
In order to correctly find a slave state including the error-bit, this line of code should be:
state = rval & 0x001f; /* read slave status */

2.	Fault Enumerations incorrect
In file ethercatprint.c the ec_ALstatuscodelist has the same AlStatusCode values for multiple faults, for example:
    {0x0035 , "DC invalid sync cycle time" },
    {0x0035 , "DC invalid sync0 cycle time" },
{0x0035 , "DC invalid sync1 cycle time" },
Therefore if an EtherCAT Slave has a ‘DC invalid sync0 cycle time’ (AlStatusCode 0x36) S.O.E.M. will print out the error string as ‘Unknown’.  
These faults should be reviewed per this website: http://infosys.beckhoff.com/english.php?content=../content/1033/ek1110/html/bt_ec_ethercat_al_statuscodes.htm&id=2342.  The three above mentioned lines should be:
    {0x0035 , "DC invalid sync cycle time" },
    {0x0036 , "DC invalid sync0 cycle time" },
{0x0037 , "DC invalid sync1 cycle time" },

3.	DC Sync functions should use signed int for parameter ‘CycleShift’
The functions ecx_dcsync0 and ecx_dcsync01 both take a parameter, CyclShift, to offset the Sync signals by a specified amount either forwards or backwards in time.  However these parameters are declared as ‘uint32’, whereas the shift time can be either positive or negative.
These functions should have CyclShift declared as a int32, not uint32

### DIFF
--- a/soem/ethercatdc.c
+++ b/soem/ethercatdc.c
@@ -8,7 +8,6 @@
  * Copyright (C) 2005-2015 Arthur Ketels
  * Copyright (C) 2008-2009 TU/e Technische Universiteit Eindhoven
  * Copyright (C) 2014-2015 rt-labs AB , Sweden
- * Copyright (C) 2015 Aversan Inc.
  *
  * SOEM is free software; you can redistribute it and/or modify it under
  * the terms of the GNU General Public License version 2 as published by the Free
@@ -69,7 +68,7 @@
  * @param [in] CyclTime         Cycltime in ns.
  * @param [in] CyclShift        CyclShift in ns.
  */
-void ecx_dcsync0(ecx_contextt *context, uint16 slave, boolean act, uint32 CyclTime, int32 CyclShift)
+void ecx_dcsync0(ecx_contextt *context, uint16 slave, boolean act, uint32 CyclTime, uint32 CyclShift)
 {
    uint8 h, RA;
    uint16 slaveh;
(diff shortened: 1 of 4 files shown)


mheden:
Thanks for your submission!

1) I can see your point, but it makes sense to let `ecx_statecheck` wait for a certain state regardless of error state. You can check the `state` variable in the `ec_slavet` structure for state + any error.

2) Good catch!

3) I can see that the comment talks about negative cycle shifts but could you describe a use case for it?

Also, remove your added copyright line in any submissions.

AversanJeff:
Making updates per comments, and creating separate pull requests for each issue

PR #19, opened 2015-12-01
Correct usage of ll in %d and %x

ShabbyX:
In 64-bit OSes, `int64_t` is just `long`. Using %lld` is wrong. This commit makes sure `%lld` always sees a `long long` regardless of architecture.

### DIFF
--- a/test/linux/slaveinfo/slaveinfo.c
+++ b/test/linux/slaveinfo/slaveinfo.c
@@ -13,6 +13,7 @@
 
 #include <stdio.h>
 #include <string.h>
+#include <inttypes.h>
 
 #include "ethercattype.h"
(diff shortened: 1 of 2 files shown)


mheden:
Thanks, however I think it would be better to use the markers found in `stdint.h`, such as `PRIi64` for this.

ShabbyX:
@mheden Feel free to do the fix then, I just wanted to bring this to attention.

ShabbyX:
@mheden I changed those cases to `PRI*64` equivalents, so this PR is ready to merge.

hefloryd:
Merged, thanks

ShabbyX:
Thank you!

PR #23, opened 2015-12-17
Allow proper installation on host system

ShabbyX:
This commit allows sane installation of SOEM on the host system, while
minimally affecting the current usage of cross-compilation.

The changes are:

- CMAKE_INSTALL_PREFIX is not overriden when installing on host,
- header files are installed in include/soem/ when installing on host,
- libsoem.a is installed in lib/ instead of bin/,
- simple_test is not installed on host.

Usage is:

    cmake -DHOST_INSTALL=1 .


hefloryd:
Merged, thanks

PR #24, opened 2015-12-18
Improve life w.r.t `#include`

ShabbyX:
Hi,

These commits make sure that a user cannot go wrong with the includes. First, it makes sure that where relative includes are needed, `""` is used instead of `<>`. Then, it creates a file that includes `nicdrv.h` and `ethercat*.h` files (except `ethercatconfiglist.h`) in the correct order, so that a user would only need to do:

    #include <ethercat.h>


ShabbyX:
So it seems that My latest two commits are also added to the pull request, so here is a brief explanation:

- The `simple_test` is now built, just not installed if doing a local install. My previous PR had it removed completely.
- The API now takes `ifname` as a `const char *` so that you can also use string literals with it.

mheden:
Thanks, will address them after the holidays.

ShabbyX:
Hi @mheden, do let me know if this PR is ok, or whether I need to make adjustments.

hefloryd:
Looks good, thanks

PR #28, opened 2016-02-01
fixing ec_ALstatuscodelist enumeration

AversanJeff:
In file ethercatprint.c the ec_ALstatuscodelist has the same AlStatusCode values for multiple faults, for example:
{0x0035 , "DC invalid sync cycle time" },
{0x0035 , "DC invalid sync0 cycle time" },
{0x0035 , "DC invalid sync1 cycle time" },
Therefore if an EtherCAT Slave has a ‘DC invalid sync0 cycle time’ (AlStatusCode 0x36) S.O.E.M. will print out the error string as ‘Unknown’.

These faults should be reviewed per this website: http://infosys.beckhoff.com/english.php?content=../content/1033/ek1110/html/bt_ec_ethercat_al_statuscodes.htm&id=2342. The three above mentioned lines should be:
{0x0035 , "DC invalid sync cycle time" },
{0x0036 , "DC invalid sync0 cycle time" },
{0x0037 , "DC invalid sync1 cycle time" },

Also added missing AlStatusCode for "MBX_AOE"

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,12 @@ cmake_minimum_required(VERSION 2.8.4)
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/Modules")
 project(SOEM C)
 
-set(CMAKE_INSTALL_PREFIX ${CMAKE_SOURCE_DIR}/install)
+if(DEFINED HOST_INSTALL)
+   set(SOEM_INCLUDE_INSTALL_DIR include/soem)
+else()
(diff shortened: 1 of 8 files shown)


ShabbyX:
There are a lot of unnecessary commits here and very little change. I suggest squashing all of this and turning it into one commit (essentially bcbf2a4 without the copy right). Or you can go back to master and do:

    $ git cherry-pick --no-commit bcbf2a4

then remove the copyright manually and commit. Then either force push to your thread, or make a new PR.

AversanJeff:
Making an new PR with fewer diffs.

PR #30, opened 2016-02-15
fixing ec_ALstatuscodelist enumeration

AversanJeff:
Fixing 'DC invalid syncX cycle time' enumerations, and adding missing
'MBX_AOE' enumeration to ec_ALstatuscodelist[].


ShabbyX:
Much nicer. This should make it simpler for the maintainers to pay attention and accept it (if it is correct, I don't know about it myself)

hefloryd:
Merged, thanks

PR #32, opened 2016-02-17
Fixed mastertime64 calculation

AversanJeff:
calculation of Master time (nanoseconds since 2000-01-01 epoch) did not correctly translate seconds into nanoseconds


AversanJeff:
I believe there's a typo in the mastertime64 calculation -- ((uint64)mastertime.sec * 1000000) will translate the 'seconds' into milliseconds, not nanoseconds

mheden:
It looks a bit silly but the code is actually correct :-) I think you are missing the last parenthesis.

First multiply `sec` with `1_000_000`, add the `usec` and then multiply the result with `1_000` to convert it to nanoseconds.

But yes, it could have be written more nicely so it doesn't look like LISP...

AversanJeff:
Ah you're right, I got confused with the brackets.  Issue closed.

PR #38, opened 2016-05-14
Rxbufstat

nakarlsson:
Improved multithreading
-Add condition to mark frame as RCVD, it must be send and marked TX
-Remove misplaced mark frame EMPTY in ecx_waitinframe, it is handled by the caller later, sr_confirm and receive processdata.
-Add mutex in InTime receive to protect RX tempbuf

### DIFF
--- a/oshw/intime/nicdrv.c
+++ b/oshw/intime/nicdrv.c
@@ -137,7 +137,13 @@ int ecx_setupnic(ecx_portt *port, const char *ifname, int secondary)
    if (status != E_OK)
    {
       ECAT_PRINT_ERROR("hpeOpen failed with status %04x ", status);
+      if(status == E_EXIST) ECAT_PRINT_ERROR("E_EXIST\n");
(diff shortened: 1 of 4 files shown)


nakarlsson:
I didn't add rxsetbuf Mutex guards in this commit, still thinking if it  is worth it. Today the only race condition situation is when looking if rxbufsat == _TX and setting it to = _RCVD, in theory the owner thread can set it _EMPTY inbetween. But for that to happen we would have a non pre-emptive scheduler, didn't feel that it is necassary. What do you think?

ArthurKetels:
The criticalsection/mutex in ecx_inframe() will take care of the race condition. If _TX to _RCVD is preempted by a higher priority task (that is also the owner of the index) the criticalsection will let the low priority task finish until after the criticalsection, so before _EMPTY is written. Because as far I can see all _EMPTY writes are preceded by ecx_inframe() without any waits or reschedueles inbetween.
Pull request merged

Issue #37, opened 2016-05-12
Grouping Slaves


BauchBeinePoe:
Basically I want to know when a part of the ethercat network is not reachable. I think groups are ment to do that?!

If I see it correctly I can group slaves by assigning a groupid to the slave but when does that need to happen? Before the `ecx_config_map_group` call or after?

Does grouping affect the mapped addresses and should I do a `ecx_send_processdata_group`  for every group or is it handled by the `ecx_send_processdata`?

Thanks in advance,
Stefan

nakarlsson:
Before ecx_config_map_group, that function make use of the groupid to make a selection if the slave should be included or not.

You have to use ecx_send_processdata_group since  ecx_send_processdata  defaults to 0.

But I wonder if it all that trouble is needed?

Why not make one group, check against expected WC, if WC isn't correct make an explicit "probe" query of a slave in that particular part?


Regards
Andreas

BauchBeinePoe:
> But I wonder if it all that trouble is needed?

This way I can simply state the applicational entity (that is associated with a subset of slaves) as not reachable and go on in the io loop instead of making extra queries. 

But back to my question, does grouping affect the adressing of the ios?

nakarlsson:
>But back to my question, does grouping affect the adressing of the ios?

No assuming a slave is only part of one group and every group have its own IOmap.

ecx_config_map_group, will map slaves belonging to given group and given IOmap making slave->outputs or slave->inputs point into given IOmap.

I believe the pseudo code below is the way to go. I'm not sure the group concept is very well tested though. Only one way to find out, please try. @ArthurKetels might have used it more.

uint8_t  IOmap[NUMBER_OF_GROUPS][SIZE];
....
for (i = 1 , i <= slave_count; i++)
    slave.group = set_group("criteria");
....

for (i = 1 , i <= NUMBER_OF_GROUPS; i++)
     ecx_config_map_group(ctx, &IOmap[i][0], i)

for (i = 1 , i <= NUMBER_OF_GROUPS; i++)
 ecx_send_processdata_group (ctx, i);

for (i = 1 , i <= NUMBER_OF_GROUPS; i++)
 ecx_receive_processdata_group(ctx, i, timeout);

BauchBeinePoe:
I did that and got a segmentation fault very often (not every time thou). Can it be fixed? That would be great!

Here is the related code. The groups have already been set.

````   
    log_debug("SEC: mapping slaves.");

[...]

PR #39, opened 2016-05-23
Fixed grouping problem.

BauchBeinePoe:
Fixed issue #37.

### DIFF
--- a/soem/ethercatmain.c
+++ b/soem/ethercatmain.c
@@ -1623,6 +1623,18 @@ static int ecx_pullindex(ecx_contextt *context)
    return rval;
 }
 
+/** 
+ * Clear the idx stack.
+ * 
+ * @param context           = context struct
+ */
+static void ecx_clearindex(ecx_contextt *context)  {
+
+   context->idxstack->pushed = 0;
+   context->idxstack->pulled = 0;
+
+}
+
 /** Transmit processdata to slaves.
  * Uses LRW, or LRD/LWR if LRW is not allowed (blockLRW).
  * Both the input and output processdata are transmitted.
@@ -1655,11 +1667,7 @@ int ecx_send_processdata_group(ecx_contextt *context, uint8 group)
    LogAdr = context->grouplist[group].logstartaddr;
    if (length)
    {
-      if(!group)
-      {
-         context->idxstack->pushed = 0;
-         context->idxstack->pulled = 0;
-      }
+
       wkc = 1;
       /* LRW blocked by one or more slaves ? */
       if (context->grouplist[group].blockLRW)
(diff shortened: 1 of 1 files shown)


nakarlsson:
I think this fix also should remove the reset made in ecx_send_processdata_group.

BauchBeinePoe:
Wouldn't that lead to fault behaviour if there are unhandled requests on the stack?

nakarlsson:
Can you describe that use-case?  I might be missing something.

BauchBeinePoe:
I guess since there is a timeout on the receive method every request is handled eventually so I was wrong, Removing the reset in the ecx_send_processdata_group should work. But I am curious why this reset is only made when the groupid is zero?

nakarlsson:
There are use-cases when you want to be able to send multiple ecx_send_processdata_group sequentially before ecx_receive_processdata_group, groups being one. I think there are use-cases that also applies to group 0 hence my suggestion to remove the unconditional reset for group 0.

I believe the reset can  be unconditional in ecx_receive_processdata_group since we for sure exit the while on the empty stack criteria, but I want @ArthurKetels opinion before we proceed.

ArthurKetels:
The change can be made unconditional. For now this is a satisfactory solution. Later I want to change ecx_receive_processdata_group() because the group argument can be removed if I add another parameter in ecx_pushindex() to flag a DC packet (indicated now with .hasdc and first). This is cleaner and solves the hackish parsing of the packet stack.

Grouped process data handling would then become:
for( i = STARTGROUP ; i <= ENDGROUP ; i++)
{
   ecx_send_processdata_group(.....);
}
.....
ecx_receive_processdata_group(.....);

Multiple send and only one receive to clean up the returned packet stack.

For good measure adding a ecx_clearindex() to reset the stack array to zero. It will make the intend more explicit.

BauchBeinePoe:
Where can one obtain the working counter for each group then?

nakarlsson:
Current design force you to handle every group explicitly (see below), the original use-case we had in mind was groups running with different update rate, in that case it don't matter. 

for( i = STARTGROUP ; i <= ENDGROUP ; i++)
 {
 ecx_send_processdata_group(.....);
 ecx_receive_processdata_group(.....);
 }

ArthurKetels:
> Where can one obtain the working counter for each group then?
That is also data that needs to be pushed on the stack, so we can attribute the WCK to the correct group. ecx_receive_processdata_group() will return the total WCK of all packets processed. The group structure wil hold the latest WCK of its own group. ecx_send_processdata_group() resets its group WKC.

This way we do not break old code that relies on current behaviour.

PR #40, opened 2016-05-26
Check if the SM seems to be faulty and no valid type

nakarlsson:


nakarlsson:
Don't enable incomplete SMs, below 2 extra SM are specifed in the EEPROM

 DCParentport:0
 Activeports:1.0.0.0
 Configured address: 1001
 Man: 00000028 ID: 000001e0 Rev: 00010000
 SM0 A:1800 L: 512 F:00010026 Type:1
 SM1 A:1a2e L: 512 F:00010022 Type:2
 SM2 A:1200 L:   6 F:00010024 Type:3
 SM3 A:1260 L:   6 F:00010000 Type:4
 SM4 A:12c0 L:   8 F:00010024 Type:0
 SM5 A:1320 L:   8 F:00010000 Type:0
 FMMU0 Ls:00000000 Ll:   6 Lsb:0 Leb:7 Ps:1200 Psb:0 Ty:02 Act:01
 FMMU1 Ls:00000006 Ll:   6 Lsb:0 Leb:7 Ps:1260 Psb:0 Ty:01 Act:01
 FMMUfunc 0:1 1:1 2:1 3:2
 MBX length wr: 512 rd: 512 MBX protocols : 04
 CoE details: 0f FoE details: 00 EoE details: 01 SoE details: 00
 Ebus current: 0[mA]

nakarlsson:
Or should they be filtered out when Reading the EEPROM, in case what critera should we use to filter them out. CtrlWord == 0?

ArthurKetels:
Actually the SM 4 and 5 are correctly defined, they just have no type assigned. You really need to have a look in the ESI xml file to know what the intention is of the 2 extra SM and 2 extra FMMU. Weird is that from the config byte SM4 would be type 3 and SM5 type 4, but the FMMU'shave type 1,1,1 and 2. So one of those two is in error. It can also be that we just see some testing remnants in the EEPROM that have been forgotten to delete.

In older slaves I have seen that they forget to put a type on the SM. With your code change we would break those slaves. 

Another problem is that the eeprom states only CoE as protocol but yet there are EoE details.

And btw, 0x200 size mailboxes are crazy too. 

Ehh, now I see, Moog is the slave vendor..... that explains a lot. I have had quite some interresting conversations with the development team in the past.

nakarlsson:
Yes they're not beeing present in the ESI file, I'm not sure my fix even solve the issue we're facing but I wanted to see if we can come up with a way to handle the issue.

There are 2 ways 

1. Make my fix work, find a way to detect a "faulty" SM and make sure we do a minimum workaround code to prevent us from failing and don't break legacy behaviuor.

Or

2. acknowledge that we can't  detect a "faulty" SM in a safe way and recommend the users to use configlist when such a case come up.

nakarlsson:
When refreshing upon how it works I see now that my fix is pretty useless, the issue would remain.

nakarlsson:
A 3:e option is to recommend the user to handled faulty SMs explicitly from the application between config_init and config_map, then they can either clear the SM or complement /correct it, depending on what is the issue.

Eg this is no SOEM issue

ArthurKetels:
In a past project of mine using Moog servo valves there where all kinds of firmware issues. Wrong eeprom content, SDO's with the wrong advertised type (UINT16 istead of UINT8), etc. Only after talking with the developer in question at Moog I could get them to update the firmware. Before the statement was : "it works with twincat".

I used the same method as you proposed (3). Do all specific configuration in a config hook routine. That way it also works when you use the hot plug and recovery functions.

nakarlsson:
After reviewing the inputs and the changes required and possible issues with a fix I agree on option 3.

hefloryd:
FYI, it was possible to fix the issue by adding a PO2SOconfig hook. See hefloryd/SOEM@22d8f5d2a23f42b795357fadfd05ea03cff49350

Issue #50, opened 2016-08-11
Thread-safety of ecx_config_map_group and its subfunctions


Apollo3zehn:
Depending on the constant **_MAX_MAPT_**, the function **_ecx_config_map_group_** calls **_ecx_map_coe_soe_** either sequentially for each slave or in parallel using **_MAX_MAPT_** threads. 

If the current slave supports complete access **_ecx_map_coe_soe_** calls **_ecx_readPDOmapCA_** and then **_ecx_readPDOassignCA_**.

The last two functions rely on the default contexts' fields **_SMcomtype_**, **_PDOassign_** and **_PDOdesc_**. These three fields seeem to be **not thread safe**, they are modified concurrently by e.g 8 threads. This leads to incorrect sync manager length calculation and then prevents the corresponding slave to get into OP.

**To solve the issue, I modified ethercatmain.c from**

```
/** SyncManager Communication Type struct to store data of one slave */
static ec_SMcommtypet   ec_SMcommtype;
/** PDO assign struct to store data of one slave */
static ec_PDOassignt    ec_PDOassign;
/** PDO description struct to store data of one slave */
static ec_PDOdesct      ec_PDOdesc;

...

ecx_contextt  ecx_context = {
...
    &ec_SMcommtype,  // .SMcommtype    =
    &ec_PDOassign,   // .PDOassign     =
    &ec_PDOdesc,     // .PDOdesc       =
...
};

```

**to** 

```
/** SyncManager Communication Type struct to store data of one slave */
static ec_SMcommtypet   ec_SMcommtype[EC_MAXSLAVE];
/** PDO assign struct to store data of one slave */
static ec_PDOassignt    ec_PDOassign[EC_MAXSLAVE];
/** PDO description struct to store data of one slave */
static ec_PDOdesct      ec_PDOdesc[EC_MAXSLAVE];

...

ecx_contextt  ecx_context = {
...
    &ec_SMcommtype[0],  // .SMcommtype    =
    &ec_PDOassign[0],   // .PDOassign     =
    &ec_PDOdesc[0],     // .PDOdesc       =
...
};

```

Additionally I replaced all occurences of SMcommtype->n to SMcommtype[Slave].n and similiar for PDOassign and PDOdesc.

Maybe someone can have a look into this.

Thank you
Vincent

nakarlsson:
@ArthurKetels, @mheden 
[...]

PR #51, opened 2016-08-17
Make PDO CA access thread safe

nakarlsson:
Fix for #50

### DIFF
--- a/soem/ethercatcoe.c
+++ b/soem/ethercatcoe.c
@@ -1452,12 +1452,13 @@ int ec_readPDOassign(uint16 Slave, uint16 PDOassign)
 /** Read PDO assign structure in Complete Access mode
  * @param[in]  Slave         = Slave number
  * @param[in]  PDOassign     = PDO assign object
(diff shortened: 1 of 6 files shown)


hefloryd:
This commit changes the license text of the affected files (e.g. Eiserstraße to Eiserstrasse). Is that intentional? If we want to change the text we should probably change all files so they are identical wrt license, and do it in a separate commit.

nakarlsson:
I did the change due to Eclipse prompting of issues about characters encoding, however you're right I shouldn't change them. I'll fix the pull request not to change the texts

hefloryd:
In Eclipse you can set UTF-8 encoding as default by going to Window -> Preferences -> Workspace -> Text file encoding.

We could also add a BOM marker to the files. However, that could theoretically cause problems with some compilers.

Also I wouldn't be opposed to changing the license text to ASCII encoding.

nakarlsson:
I reverted the licese change in a new commit, didn't figure out of to undo specific lines only.

Issue #60, opened 2016-09-08
ecx_dcsync01: slave fails going to OP with 0x002C: fatal sync error


Apollo3zehn:
Since I make use of the functions **ecx_dcsync0** and **ecx_dcsync01**, the terminals EL3104 and (if I remember correctly) also EL3154 fail to go into OP with error 0x002C: fatal sync error. This always happens when using one of the OpMode "**DC-Synchron**" or "**DC-Synchron (input based)**", which are specified in the ESI XML file. 

The first time I start the system everything works, but the second time (and third and so on) it fails. Using wireshark I noticed that TwinCAT resets register 0x0981 long before the DC system is configured. I think this corresponds to the function ecx_set_slaves_to_default. I included the following code

_static void ecx_set_slaves_to_default(ecx_contextt *context)
{
...
b = 0x00;
ecx_BWR(context->port, 0x0000, ECT_REG_DCSYNCACT, sizeof(b), &b), EC_TIMEOUTRET3);
...
}_
and since then everything works fine. Is somebody able to reproduce this?

Thanks
Vincent

ArthurKetels:
Nice catch. It does make sense to reset sync generation at early config. On the other hand a slave should not complain when sync pulses are generated in states before OP.

Apollo3zehn:
One correction: Sometimes it failed to get into OP as described above, and sometimes it went into OP but did not deliver process data.

nakarlsson:
@Apollo3zehn , can you create a pull request?

nakarlsson:
Pull request commited

PR #64, opened 2016-12-07
fix issue #60 (slave fails going to OP: fatal sync error)

Apollo3zehn:
- reset activation register in ecx_set_slaves_to_default

### DIFF
--- a/osal/win32/osal.c
+++ b/osal/win32/osal.c
@@ -48,6 +48,24 @@ static double qpc2usec;
 
 #define USECS_PER_SEC     1000000
 
+int osal_getabsolutetimeofday(struct timeval *tv, struct timezone *tz)
+{
+	FILETIME SystemTime;
+	int64 SystemTime64, usecs;
+	int64 Offset = -134774LL * 86400LL * 1000LL * 1000LL * 10LL; // Difference between 1601 and 1970 in 100 ns.
+
+	GetSystemTimeAsFileTime(&SystemTime);
+
+	SystemTime64 = ((int64)(SystemTime.dwHighDateTime) << 32) + (int64)SystemTime.dwLowDateTime;
+	SystemTime64 = SystemTime64 + Offset;
+	usecs = SystemTime64 / 10;
+
+	tv->tv_sec = (long)(usecs / 1000000);
+	tv->tv_usec = (long)(usecs - (tv->tv_sec * 1000000));
+
+	return 1;
+}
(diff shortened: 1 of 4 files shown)


soem/ethercattype.h
```diff
@@ -86,9 +86,9 @@ extern "C"
 /** datagram type EtherCAT */
 #define EC_ECATTYPE        0x1000
 /** number of frame buffers per channel (tx, rx1 rx2) */
-#define EC_MAXBUF          16
+#define EC_MAXBUF          256
```
nakarlsson: Leave ethercattype.h as is, changes here affects all plattforms. I've written an #65 to make it possible to override values.  256 buffers don't make sense on small embedded systems.

osal/win32/osal.c
```diff
@@ -48,6 +48,24 @@ static double qpc2usec;
 
 #define USECS_PER_SEC     1000000
 
+int osal_getabsolutetimeofday(struct timeval *tv, struct timezone *tz)
```
nakarlsson: How does this function get affected by system time?  For the stack the most important task is to work as a stop watch and monitor when we have waited or tried long enough. It is kind of legacy from Linux that it is named "gettimeofday".

Apollo3zehn:
Sorry, I thought that only my initial commit will be included in the pull request. The following commits were indeed only intended for my fork.

Is it better to recreate this pull request or are you able to deny all changes except those in ethercatconfig.c ?

osal/win32/osal.c
Apollo3zehn: This function provides a new system time value every 1-15 ms (depending on the settings made with [timeBeginPeriod](https://msdn.microsoft.com/en-us/library/windows/desktop/dd757624(v=vs.85).aspx)). On Windows 8 and above, there is the new function [GetSystemTimePreciseAsFileTime](https://msdn.microsoft.com/en-us/library/windows/desktop/hh706895(v=vs.85).aspx) that provides the system time with much higher precision (< 1 µs). 

My version of osal_getabsolutetimeofday is only a workaround to initialize DC with an absolute time but needs some modifications before publishing. I would prefer to create a separate pull request for this issue. As I wrote below, I mistakenly commited serveral changes to this pull request.

nakarlsson:
Please recreate the pull request.

Apollo3zehn:
created new pull request with correct changes

Issue #22, opened 2015-12-10
Linux: CMAKE_INSTALL_PREFIX overriden in CMakeLists.txt


ShabbyX:
`CMAKE_INSTALL_PREFIX` being set to `source/install`, it is impossible to override this in command line to actually install the library in the system. Would you mind if I remove it and send a PR?

hefloryd:
Our main use-case is to cross-compile SOEM for other architectures, where it makes little sense to install it in the host system. But it would be easy enough to explicitly configure it that way if that is required. Please send a PR if you like, possibly defaulting to not install in the host system.

ShabbyX:
Ok, I see. I'd do that, so you wouldn't have to change anything, but it would also be possible to install to host!

hefloryd:
Sounds good.

PR #84, opened 2017-03-10
Allow out-of-source builds

alberth:
CMake supports out-of-source builds, which is normally done like
```` sh
mkdir my_build_dir
cmake ..   # Point back to SOEM root
make
make install
````
For SOEM this fails, as it hard-codes the install directory to be SOEM/install.

This patch fixes that by setting the install prefix to the binary root (ie /abs/path/to/SOEM/my_build_dir above), except in the case of "cmake ." where it switches back to the old behaviour to avoid making a mess in SOEM root for existing users.

PS CMake tends to litter generated files all over the place, and these tend to take precedence over new CMake runs, so when testing, make very sure you don't have old generated files laying around that mess up the build / install.


hefloryd:
See #22 and the discussion there. You can install into the host system by setting -DHOST_INSTALL=1.

alberth:
I wasn't doing a proper system install, nor is #22 doing an out-of-source build.

The difference is that an out-of-source build with CMake dumps all the CMake generated files in the out-of-source build dir, instead of all over the place in the SOEM source tree.
(It's an assumption of CMake that you do this, this is why CMake (unlike eg autotools) doesn't provide a "remove all generated files" target, it assumes you can do ``rm -r my_build_dir``.)

But fair enough, if setting a non-standard variable is required to get default CMake-intended behaviour, so be it.

----
One thing I found with my patch while further experimenting with your answer, is that my patch cannot do custom install locations (I didn't notice that at the time as I wasn't installing). It shouldn't touch the CMAKE_INSTALL_PREFIX, so it continues to be user-settable.

hefloryd:
Out-of-source builds is the recommended way of building SOEM, see the README.md. 

$ mkdir build
$ cd build/
$ cmake ..
$ make

Is this not working for you? At this point if you do a make install it will install into one isolated folder in the parent directory. No need to build in tree.

alberth:
Out of source build includes not dumping results in a fixed in-source directory for me.
```` bash
mkdir ../build_dir
cd ../build_dir
cmake -DCMAKE_INSTALL_PREFIX=./installed ../SOEM
make # Does not add files to ../SOEM
make install # Dumps results in ../SOEM/install instead of ./installed
````
The final ``make install`` doesn't honour my CMAKE_INSTALL_PREFIX request for an out-of-source build.

I agree "-DCMAKE_INSTALL_PREFIX=<path>" is pretty horrible to type, the default ``/usr/<somewhere>`` isn't very useful for testing/debugging, and that SOEM/install is a pretty good point to dump build-results while developing and debugging SOEM.

However, why do you rig cmake to the point it ignores requests as defined in the cmake manual, and then add a non-standard setting to revert back to original cmake behavior, if you can also add a simple ``setup_soem_development.sh`` script in the root of the project like
```` bash
#!/bin/sh
cmake -DCMAKE_INSTALL_PREFIX=./install .
````

But perhaps I just don't understand what you're trying to do here, or we disagree on what an out-of-source build is exactly, or we disagree on badness of ignoring what cmake manuals say. Whatever the reason, I have now
```` sh
#!/bin/sh
set -e -u -x

if test -d installed_soem
then rm -rf installed_soem
fi

if test -d build_dir
then rm -rf build_dir
fi

mkdir build_dir
cd build_dir
cmake -DHOST_INSTALL=1 \
      -DCMAKE_INSTALL_PREFIX=../installed_soem \
      -DCMAKE_C_COMPILER:FILEPATH=gcc \
      -DCMAKE_C_FLAGS:STRING=-fPIC \
      ../SOEM

make
make install

cd ..
rm -r build_dir
````
that builds the library for me at the spot where I want it, in the way that I want it.

hefloryd:
It seems that your problem is that you can't set CMAKE_INSTALL_PREFIX. Please check if PR #86 works for you.

PR #86, opened 2017-03-13
allow setting CMAKE_INSTALL_PREFIX externally

hefloryd:
This lets the user specify CMAKE_INSTALL_PREFIX on the command line. Also removes -DHOST_INSTALL as you can specify -DCMAKE_INSTALL_PREFIX=/usr/local instead. Note that the include install path is now always include/soem.


alberth:
Works

Thanks for your efforts.

hefloryd:
Thanks, I removed the HOST_INSTALL reference from simple_test (so it will always be installed).

PR #87, opened 2017-03-14
nicdrv.c changes for VxWorks 6.9 target

HolgerNie:
As I learned from WRS for some old network cards it works without to set this but for the new one we need also to set mBlkPktHdr.len. Their advice is to set also this filed for mBlk structure when muxSend is used.


nakarlsson:
Could you please squash it to one commit.

HolgerNie:
Sorry, but I have no idea how to change this

nakarlsson:
thanks

PR #89, opened 2017-03-15
CMake: Add option to build shared library

thopiekar:
Just added a new variable "BUILD_STATIC" to build libsoem statically or not.
By default it will now build libsoem as shared library.


hefloryd:
Our main use-case is to cross-compile SOEM for other architectures, most of which don't have shared libraries. An easy way to enable shared libraries would be useful but it should default to static libraries.

Also, did you try building this on windows? It seems to fail there.

thopiekar:
Hey @hefloryd ,

first of all thank you for your feedback. I personally work in a pure Linux environment and already packaged here locally SOEM as an debian package. For me it makes sense to build it as an shared library and so far it works perfectly.
Never tested my work on other OSs, also if you like that SOEM shall be build statically by default I can change my PR for this.

I'm open for any discussion 😉

jumo-sballing:
Building SOEM as shared lib would be a great feature. I have to edit the makefile manually every time. 
Unfortunatelly I also cannot test the build on windows.

PR #106, opened 2017-05-23
Fixed osal_timer_diff bug

zchen24:
ec_timet->usec is uint32 type and will never be smaller than 0


nakarlsson:
Good catch, can you please change all issues? Linux,Win...

zchen24:
Of course. The commit has been updated. Thanks.

PR #116, opened 2017-07-12
Added Xenomai support

zchen24:
Hi,

The pull request contains my port of this library to support Xenomai. This port has been tested on an EtherCAT slave device from TI (AM3359 Industrial Communications Engine) using ```simpletest``` and ```simpleinfo```. I was able to do cyclic data exchange.

The port adds two CMake modules files to the ```cmake``` folder. One is for Xenomai, which is from ROS cmake_modules (https://github.com/ros/cmake_modules/blob/0.4-devel/cmake/Modules/FindXenomai.cmake) and originally from JHU (cisst license, BSD-like). The other one is for RTNet and from https://github.com/kuka-isir/ati_sensor/blob/master/cmake/Modules/FindRTnet.cmake (GPL 2.0). I created separate xenomai folders in ```osal``` and ```oshw```, copied Linux code and made some changes. I noticed that the difference between the xenomai and linux osal / oshw code is minimum (less than 10 lines). Therefore, another option is to add conditional compilation to the Linux source code directly. Please let me know your preference. It does require some setup to use SOEM on Xenomai and RTNet. I can write a brief how-to if it is necessary.

It does require some setup to use SOEM on Xenomai and RTNet. I can write a brief how-to if it is necessary. 

Thanks, 
Zihan


nakarlsson:
Are the changes in the test/linux source compatible with Linux or are they Xenomai specific?

hefloryd:
If FindRTnet.cmake is GPL2 then we may not be able to use it, since SOEM is licensed as GPL2 with an exception clause. The file itself doesn't contain any license information however. Can you check with the RTnet developers if it is OK for us to use the file?

Another way around it may be to assume the file is available externally so that we don't need to distribute it with SOEM.

Regarding the modifications to test/linux source: it is not obvious but for a while now we use those sources to build for windows and other OS:es as well. We should really clean up the test folder. However, the changes this pull request makes to these files will currently break builds for other OS:es.

zchen24:
Hi Andreas, Good catch, I need to do a test and will let you know.

zchen24:
@hefloryd The author replied: 
"No problem for this file..it's from somewhere else if I recall correctly. Happy coding !"

The email conversion has been forwarded to @nakarlsson and support@rt-labs.com.
Is this good enough?

muttistefano:
Hello,
i'm trying to make this work but FindRTnet.cmake cannot find Rtnet.
I am using Xenomai 3.0.5 and if i'm not wrong i undestrood that Rtnet is included in Xenomay 3 without the need to install it.
Any idea?
Thanks

claudioscordino:
Hi.
I wonder if there will be any follow-up of this pull request...
Having hard-RT support for SOEM would be quite interesting.

BrettHemes:
Any comment as to why this was closed?

muenchenstrasse:
> @hefloryd The author replied:
> "No problem for this file..it's from somewhere else if I recall correctly. Happy coding !"
> 
> The email conversion has been forwarded to @nakarlsson and [support@rt-labs.com](mailto:support@rt-labs.com).
> Is this good enough?

Hello, is the newest CMAKE from SOEM appended to apply for RTNET and Xenomai?

PR #123, opened 2017-08-15
Enable and fix warnings

hefloryd:
Cleanup CMakeLists.txt and enable GCC warnings -Wall and -Wextra. Fix some warnings that this produces. Also enable -Werror to avoid new warnings.


hefloryd:
Removed line-endings fix, will post new PR

nakarlsson:
could you remove old Makefiles as a bonus in this PR?

Issue #144, opened 2018-01-09
ec_slave[slave].state operational when slave is lost


kajusK:
Hi,
I've been fighting with the simple_test example on windows. There are several issues when running setup with several slaves:

- expected WKC calculation - when using 2 slaves (with both inputs and outputs configured), expected WKC is calculated as 6, real WKC returned during communication is 5. I'm pretty new to ethercat, but shouldn't be the correct calculation of expected wkc like:
```C
 (ec_group[0].outputsWKC*2 - 1)+ec_group[0].inputsWKC;
```

Anyway, that's not a blocker, the more serious issue I hit is state when a slave is lost (disconnected). Error detection works as expected when I disconnect slave directly connected to the master, but when I disconnect a different slave, the ec_slave[slave].state remains EC_STATE_OPERATIONAL for all connected slaves, any idea why? simple_test starts printing `OK : all slaves resumed OPERATIONAL.`

Thanks,
Jakub

nakarlsson:
About WKC, check chapter 2.4 in https://download.beckhoff.com/download/document/io/ethercat-development-products/ethercat_esc_datasheet_sec1_technology_2i3.pdf. I guess you execute LRW and would get a read & write +3 giving 2 *3 = 6.

Are you running linux/simple_test/simple_test.c?

nakarlsson:
Could you provide a console log snippet and wireshark log when it happens?

kajusK:
Hi,
aha, I found the issue with wkc, the second slave was not initialized correctly, now it works as expected. Thank you.

Yes, I'm running linux/simple_test/simple_test.c. Although I running on windows due to the project requirements (I'm linux only user otherwise), the linux version works just fine. I've also tried to compile win32 version and run simple_test again, but with the same result.

The requested logs are below. First, the second slave was disconnected, reconnected and then the same procedure happened with the first slave.
 
[slaveinfo.txt](https://github.com/OpenEtherCATsociety/SOEM/files/1621754/slaveinfo.txt)
[simple_test.txt](https://github.com/OpenEtherCATsociety/SOEM/files/1621757/simple_test.txt)
[wireshark.zip](https://github.com/OpenEtherCATsociety/SOEM/files/1621758/wireshark.zip)

nakarlsson:
[...]

PR #155, opened 2018-01-31
Add check that all slaves are present

nakarlsson:
Tested in #144 , we need to check if all slaves are present, otherwise we might think everything is OK since every slave responding is in OP without error, but we've actually lost as slave that trigger the error checking

### DIFF
--- a/soem/ethercatmain.c
+++ b/soem/ethercatmain.c
@@ -724,11 +724,19 @@ int ecx_readstate(ecx_contextt *context)
    ec_alstatust sl[MAX_FPRD_MULTI];
    uint16 slca[MAX_FPRD_MULTI];
    boolean noerrorflag, allslavessamestate;
+   boolean allslavespresent = FALSE;
+   int wkc;
 
    /* Try to establish the state of all slaves sending only one broadcast datargam.
     * This way a number of datagrams equal to the number of slaves will be sent only if needed.*/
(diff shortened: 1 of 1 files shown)


nakarlsson:
@MagnaboscoL , seems OK with you?

MagnaboscoL:
Ok for me, thanks for the bugfix!

soem/ethercatmain.c
```diff
@@ -724,11 +724,19 @@ int ecx_readstate(ecx_contextt *context)
...
+   boolean allslavespresent = FALSE;
+   int wkc;
 
    /* Try to establish the state of all slaves sending only one broadcast datargam.
     * This way a number of datagrams equal to the number of slaves will be sent only if needed.*/
    rval = 0;
-   ecx_BRD(context->port, 0, ECT_REG_ALSTAT, sizeof(rval), &rval, EC_TIMEOUTRET);
+   wkc = ecx_BRD(context->port, 0, ECT_REG_ALSTAT, sizeof(rval), &rval, EC_TIMEOUTRET);
+
+   if(wkc == *(context->slavecount))
```
ArthurKetels: Should make == into >=. There can be the case that an extra slave is added to the network. So wkc can be larger than slavecount

nakarlsson: Will do

PR #156, opened 2018-02-01
Add functions to configure and send an overlapping IOmap

nakarlsson:
Main objectiv: Workaround for TI ESC errata  – SDOCM00105048/PINDSW-141: LRW access to non-interleaved input and output process data of multiple slaves does not work. SOEM accesses slaves in LRW mode this way

Also: Using an overlapped LRW frame get smaller per definition

### DIFF
--- a/soem/ethercatconfig.c
+++ b/soem/ethercatconfig.c
@@ -820,7 +820,309 @@ static int ecx_get_threadcount(void)
    return thrc;
 }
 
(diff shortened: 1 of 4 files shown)


nakarlsson:
PDF over overlapping IOmap layout

[overlapping_IOmap.pdf](https://github.com/OpenEtherCATsociety/SOEM/files/1685407/overlapping_IOmap.pdf)

PR #157, opened 2018-02-02
When broadcast read the state,it is OK if the WKC is higher

nakarlsson:
In addition to previous commit, https://github.com/OpenEtherCATsociety/SOEM/pull/155.


nakarlsson:
Already reviewed

PR #187, opened 2018-07-03
Rtems5 integration

lounick:
Some minor nits as I was testing with RTEMS 5.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,7 @@ endif()
 
 set(SOEM_INCLUDE_INSTALL_DIR include/soem)
 set(SOEM_LIB_INSTALL_DIR lib)
+set(BUILD_TESTS TRUE)
 
 if(WIN32)
   set(OS "win32")
@@ -41,6 +42,7 @@ elseif(${CMAKE_SYSTEM_NAME} MATCHES "rtems")
   message("Building for RTEMS")
   set(OS "rtems")
   set(SOEM_LIB_INSTALL_DIR ${LIB_DIR})
+  set(BUILD_TESTS FALSE)
 endif()
 
 message("OS is ${OS}")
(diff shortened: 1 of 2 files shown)


nakarlsson:
can you squash the commits?

lounick:
Sorry forgot about it :smile:

CMakeLists.txt
```diff
@@ -74,6 +74,8 @@ install(FILES
   ${OSHW_HEADERS}
   DESTINATION ${SOEM_INCLUDE_INSTALL_DIR})
 
+if(NOT ${CMAKE_SYSTEM_NAME} MATCHES "rtems")
```
hefloryd: Is it not possible to build these on rtems? They're often useful as a quick sanity-check of SOEM and the network.

lounick: I'm getting some linking errors. I could introduce linking against libm on RTEMS targets as it doesn't find `log10`.

hefloryd: Sounds like a good idea, if needed on rtems. (I don't think log10 is used on other platforms?)

lounick: Hmmmm for some reason even if I try to link I get an error. I'll investigate but it will have to wait for tomorrow.

lounick: @hefloryd @nakarlsson any news on this? I don't have more time to investigate and it would block building SOEM for RTEMS users. So I can't push the relevant patches on the other side.

lounick: In RTEMS they use waf as a build system to find the correct dependencies. See: [soem-rtems-wscript](https://github.com/Intermodalics/soem-rtems/blob/master/wscript), [slaveinfo-wscript](https://github.com/Intermodalics/soem-rtems/blob/master/tests/slaveinfo/wscript), [rtems-waf](https://github.com/RTEMS/rtems_waf). This allows to build successfully the [slaveinfo](https://github.com/Intermodalics/soem-rtems/blob/master/tests/slaveinfo/slaveinfo.c) that has to be modified as well.

Correct that a tests flag should be introduced that will default to true and in rtems targets will be set to false. It is cleaner that way.

Issue #141, opened 2017-12-21
Check application timeouts


nakarlsson:
Check application timeouts

PR #189, opened 2018-07-27
Allow seting the desired end state in ecx_config_init

jespersmith:
This PR adds a new function ecx_config_init_to_state which adds the final desired state to ecx_config_init. Valid states would be INIT, BOOT and PRE_OP.

ecx_config_init calls ecx_config_init_to_state with PRE_OP as desired state.

This PR makes it possible to go directly from INIT to BOOT to re-attempt a firmware update. In our slave, the bootloader will exit once the slave is requested to go to PRE_OP. Should a firmware upgrade fail, there is no way to go the BOOT mode if the ecx_config_init transitions the slave to PRE_OP before then.

### DIFF
--- a/soem/ethercatconfig.c
+++ b/soem/ethercatconfig.c
@@ -306,9 +306,10 @@ static int ecx_lookup_prev_sii(ecx_contextt *context, uint16 slave)
  *
  * @param[in] context      = context struct
  * @param[in] usetable     = TRUE when using configtable to init slaves, FALSE otherwise
+ * @param[in] state        = Desired state after init. Valid options are EC_STATE_INIT, EC_STATE_BOOT or EC_STATE_PRE_OP
  * @return Workcounter of slave discover datagram = number of slaves found
  */
-int ecx_config_init(ecx_contextt *context, uint8 usetable)
+int ecx_config_init_to_state(ecx_contextt *context, uint8 usetable, ec_state state)
 {
    uint16 slave, ADPh, configadr, ssigen;
    uint16 topology, estat;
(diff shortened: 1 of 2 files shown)


nakarlsson:
I like the idea and I think the same should apply to config_map. Optional to enter SAFE_OP. But for the current API adding to_state will make the function names realy long. Any idea of good naming? Overlap might need to be refactored.

jespersmith:
I can probably adjust config_map if I find time. It's marked legacy, haven't investigated the replacement yet though.

I'm mostly a Java developer so long names are normal to me. 
If for backwards compatibility overlap should be avoided, I can think of the following options: 

- ecx_config_init_state
- ecx_config_init2
- ecx_config_to_state

For ecx_config_map_group it can become
- ecx_config_map_group_state
- ecx_config_map_group2
- ecx_config_map_to_state

nakarlsson:
I might have confused you a bit, by overlap I meant **overlap** as in the function namn **ec_config_overlap_map_group** , this will become quite long if you add the to_state.
Overlap means the overlay of inputs/outputs in the processdata.
ec_config_map_group_to_state is OK but with overlap in the name it becomes realy long. ec_config_overlap_map_group_to_state.

Perhaps best todo config map in another PR

jespersmith:
Agree, I'll keep the PR small and as is then.

nakarlsson:
I’m bit not sure how to handle the possibillity to try to enter an illegal state, have you though about that?

Would it make sens to create explicit alt.

.._init

..._init_to_preop
..._init_to_boot
..._init_to_init

jespersmith:
There is no error checking currently, passing in SAFE_OP or OP will probably result in an invalid state. 

Options)
A) I could return wkc = 0 if the state is not EC_STATE_INIT, EC_STATE_BOOT or EC_STATE_PRE_OP.
B) Explicit might be nice, it does add a few extra functions which would increase maintenance a bit. 

Let me know what you prefer and I'll update the PR.

nakarlsson:
I can’t settle for a perfect solution, best thing, in my opinion, would be to remove automatic state chaning in mentioned functions and leave that to the user/master developer. The API would remain intact but functions config_init and config_map would not change state. The stats change is an undocumented feature.



What do you think @ArthurKetels?

ArthurKetels:
I think we can not break existing code. In hindsight it would have been better to have no state change in the init, config and map functions.

I think an acceptable way around is to make an extra flag variable in the context struct (call it auto_state_change) make it 1 by default. The config an map functions then query and execute according this flag. We still do the checks to prevent execution in the wrong state.

We make some helper functions for standard state changes, ecx_init_to_boot, exc_init_to_preop, ecx_preop_to_safeop etc. Also helpers like ecx_waitfor_boot(timeout), ecx_waitfor_preop(timeout) etc.

nakarlsson:
I respect that, basically all master applications would break.
Also, the helper functions _waitfor_ implemented in the samples would  fix https://github.com/OpenEtherCATsociety/SOEM/issues/141 that Reiner addressed.

nakarlsson:
@jespersmith , I guess it would work for you aswell? Flag to stop the state change in config_init and an application based state change?

jespersmith:
@nakarlsson That would work for me. If you want I can take a shot at making the change, but I don't have much time the coming weeks.

Any issue with breaking binary compatibility on the context struct. Or just add at the end?

nakarlsson:
> Any issue with breaking binary compatibility on the context struct. Or just add at the end?
Don't know to be honest, I'm not aware of such usage but you never know. 
On the other hand I see no reason why it can't be added at the end.

mwh-newtec:
@jespersmith Did you ever implement the above discussed solution?
We have made exactly the same solution as your initial proposal, however the solution discussed above seems "nicer". So just wanted to hear you out on your status.

jespersmith:
@mwh-newtec I won't have time to implement it till February unfortunately. It's still on my task list and I'll make a PR when I get around it.

PR #200, opened 2018-08-22
Fix race condition in rxbufstat

Wanga:
I skip return checking in oshw/rtk because there already have ASSERT inside bfin_EMAC_send


nakarlsson:
Why is the mutex prio commit included in this PR?

Wanga:
please check again

nakarlsson:
Looks good, thanks for the patch

PR #215, opened 2018-10-01
Adds BUILD_SHARED_LIBS option

carlosjoserg:
It is basically the same as https://github.com/OpenEtherCATsociety/SOEM/pull/89 by @thopiekar but configured opt-out to keep the current default build procedure as static and not break pipelines around

It also provides feedback about the install prefix, since it is usually `/usr/local` and not `CMAKE_SOURCE_DIR`

Suggestions welcome!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,10 +6,14 @@ if (CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
   # Default to installing in SOEM source directory
   set(CMAKE_INSTALL_PREFIX ${CMAKE_CURRENT_LIST_DIR}/install)
 endif()
+message("CMAKE_INSTALL_PREFIX is ${CMAKE_INSTALL_PREFIX}")
 
 set(SOEM_INCLUDE_INSTALL_DIR include/soem)
 set(SOEM_LIB_INSTALL_DIR lib)
 set(BUILD_TESTS TRUE)
+
+option(BUILD_SHARED_LIBS "Build shared libraries" OFF)
+message("BUILD_SHARED_LIBS is ${BUILD_SHARED_LIBS}")
 
 if(WIN32)
   set(OS "win32")
@@ -60,12 +64,26 @@ file(GLOB SOEM_HEADERS soem/*.h)
 file(GLOB OSAL_HEADERS osal/osal.h osal/${OS}/*.h)
 file(GLOB OSHW_HEADERS oshw/${OS}/*.h)
 
-add_library(soem STATIC
-  ${SOEM_SOURCES}
-  ${OSAL_SOURCES}
-  ${OSHW_SOURCES}
-  ${OSHW_EXTRA_SOURCES})
-target_link_libraries(soem ${OS_LIBS})
+include_directories(soem)
+include_directories(osal)
+include_directories(osal/${OS})
+include_directories(oshw/${OS})
+
+if(BUILD_SHARED_LIBS)
+  add_library(soem SHARED
+    ${SOEM_SOURCES}
+    ${OSAL_SOURCES}
+    ${OSHW_SOURCES}
+    ${OSHW_EXTRA_SOURCES})
+  target_link_libraries(soem ${OS_LIBS})
+else()
+  add_library(soem STATIC
+    ${SOEM_SOURCES}
+    ${OSAL_SOURCES}
+    ${OSHW_SOURCES}
+    ${OSHW_EXTRA_SOURCES})
+    target_link_libraries(soem ${OS_LIBS})
+endif()
 
 target_include_directories(soem PUBLIC
   $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/soem>
@@ -96,7 +114,7 @@ install(FILES
   ${OSHW_HEADERS}
   DESTINATION ${SOEM_INCLUDE_INSTALL_DIR})
 
-if(BUILD_TESTS) 
+if(BUILD_TESTS)
   add_subdirectory(test/linux/slaveinfo)
   add_subdirectory(test/linux/eepromtool)
   add_subdirectory(test/linux/simple_test)


CMakeLists.txt
```diff
@@ -60,12 +64,23 @@ include_directories(osal)
 include_directories(osal/${OS})
 include_directories(oshw/${OS})
 
-add_library(soem STATIC
-  ${SOEM_SOURCES}
-  ${OSAL_SOURCES}
-  ${OSHW_SOURCES}
-  ${OSHW_EXTRA_SOURCES})
-target_link_libraries(soem ${OS_LIBS})
+
```
hefloryd: Remove added whitespace

CMakeLists.txt
```diff
@@ -60,12 +64,23 @@ include_directories(osal)
...
+
+if(BUILD_SHARED_LIBS)
+  add_library(soem SHARED
+    ${SOEM_SOURCES}
+    ${OSAL_SOURCES}
+    ${OSHW_SOURCES}
+    ${OSHW_EXTRA_SOURCES})
+  target_link_libraries(soem ${OS_LIBS})
```
hefloryd: target_link_libraries is independent of BUILD_SHARED_LIBS

hefloryd: set_property(TARGET soem PROPERTY POSITION_INDEPENDENT_CODE ON)?

CMakeLists.txt
```diff
@@ -60,12 +64,23 @@ include_directories(osal)
...
 
-add_library(soem STATIC
-  ${SOEM_SOURCES}
-  ${OSAL_SOURCES}
-  ${OSHW_SOURCES}
-  ${OSHW_EXTRA_SOURCES})
-target_link_libraries(soem ${OS_LIBS})
+
+
+if(BUILD_SHARED_LIBS)
```
hefloryd: Is this needed? Seems like it would suffice to just set BUILD_SHARED_LIBS to TRUE.
https://cmake.org/cmake/help/v3.0/variable/BUILD_SHARED_LIBS.html#variable:BUILD_SHARED_LIBS

CMakeLists.txt
hefloryd: Ignore that, seems -fPIC is set automatically when building SHARED so this is not needed.

CMakeLists.txt
carlosjoserg: Yep, sorry about that

CMakeLists.txt
carlosjoserg: You had `add_library(soem STATIC`

If the word STATIC is removed, then yes, it suffices to set that option to TRUE, it is global

carlosjoserg:
> This doesn't work on Windows? It would be great if you could fix that. Alternatively, maybe add a message and abort build if on windows and BUILD_SHARED_LIBS is set.

I don't know, but it should, what it's added does not have anything to do with the OS, as far as I am concerned

CMakeLists.txt
carlosjoserg: Copied

carlosjoserg:
There is an alternative, and it is to build both using something like:

``` cmake
add_library(soem STATIC
  ${SOEM_SOURCES}
  ${OSAL_SOURCES}
  ${OSHW_SOURCES}
  ${OSHW_EXTRA_SOURCES})

add_library(soem_shared SHARED
  ${SOEM_SOURCES}
  ${OSAL_SOURCES}
  ${OSHW_SOURCES}
  ${OSHW_EXTRA_SOURCES})
target_link_libraries(soem_shared ${OS_LIBS})
set_target_properties(soem_shared PROPERTIES
  OUTPUT_NAME soem)
```
Installing a renamed target works as well, for instance this:
``` cmake
install(TARGETS soem_shared DESTINATION ${SOEM_LIB_INSTALL_DIR})
```
copies the `libsoem.so` to the install dir.

@hefloryd   Does that look like better way? Is there any inconveniente to have both static and shared libraries installed?  If so, I can close this PR and prepare another one (considering your comments here as well)

hefloryd:
> @hefloryd Does that look like better way? Is there any inconveniente to have both static and shared libraries installed? If so, I can close this PR and prepare another one (considering your comments here as well)

I don't think there is any inconvenience really but it would probably break the windows build (this PR doesn't work there, see https://ci.appveyor.com/project/hefloryd/soem/build/23). It may also break other (embedded) targets that don't support shared libraries.

I would suggest just adding the BUILD_SHARED_LIBS option, defaulting to OFF, and removing the word STATIC. Looks like that would be enough since cmake seems to handle the rest automatically.

carlosjoserg:
I tried the approach [above](https://github.com/OpenEtherCATsociety/SOEM/pull/215#issuecomment-426209825) building both static and shared simultaneously.. using appveyor results in a successful build, check it out:
https://ci.appveyor.com/project/carlosjoserg/soem-1

hefloryd:
That's interesting. I am not sure but I think there may be a problem with the build - you might be overwriting the static library with the import library for the dll, hence the "up-to-date" message for soem.lib in the install section (if cmake decides based on timestamps, if they are actually identical it wouldn't matter of course. I don't know windows well enough to say for sure).

Also, what would happen with targets that don't support shared libraries? I don't see how cmake would be able to proceed with the shared lib section (unless it just skips it). Just adding the BUILD_SHARED_LIBS option seems like a safer change.

carlosjoserg:
Hey, I guess there are no too many windows users watching this repo to give a hand on the matter

Just read a bunch of posts around, some of them mentioning the [DLL hell](https://en.wikipedia.org/wiki/DLL_Hell), to finally agree: it is safer to add an option default to OFF

Probably the most interesting post, if anyone is interested:
http://gernotklingler.com/blog/creating-using-shared-libraries-different-compilers-different-operating-systems/

I'll update the PR with your comments then, thanks

carlosjoserg:
Hi there, it's been a while.

I'm not completely sure of my own PR after two years... :)

We definitely use it in shared mode, but I guess it's static also to have a tiny bit more of runtime performance and avoid `ld` weird configurations.

So, I'm gonna close this PR, and actually consider to link statically this lib.

Thanks!

Issue #174, opened 2018-06-01
Endianness problem on PPC604 target


willytax:
I'm trying to use SOEM on a PPC604 target (MPC5200) running busybox (linux kernel 2.6.28), therefore in ethercattype.h I changed #define EC_LITTLE_ENDIAN to #define EC_BIG_ENDIAN.
I compiled the slaveinfo example code on Eclipse using a cross-compiler. 
Comparing the output that I get on the ppc machine with the one I obtain running the same code on a linux intel PC I can see that tsome endianness problems persist. For example all the addresses displayed are byte swapped and I can't get the device map correctly.
I'm missing something?
Has anyone encountered this problem yet?

Thankyou

nakarlsson:
I’m sure we have more run-time on Little endian and I’m not sure how many installations running on Big Endian at all. So what you see is what you get, feel free to lift the Big endian version of SOEM or if we’re lucky someone have the missing pieces in a patch ready.

We at rt-labs have never run SOEM on Big Endian.

nakarlsson:
Can you give some more details round this issue?

nakarlsson:
Fixes have been done in Endian fixes #222 , write a new issue if more fixes is needed

PR #222, opened 2018-10-30
Endian fixes

Schlumpf:
As @willytax mentioned in #174, there are some problems on big-endian systems. I also use SOEM on a big-endian PowerPC system and found a few bugs. No guarantee, that there aren't more.

### DIFF
--- a/soem/ethercatconfig.c
+++ b/soem/ethercatconfig.c
@@ -166,8 +166,8 @@ static void ecx_set_slaves_to_default(ecx_contextt *context)
    ecx_BWR(context->port, 0x0000, ECT_REG_RXERR       , 8         , &zbuf, EC_TIMEOUTRET3);  /* reset CRC counters */
    ecx_BWR(context->port, 0x0000, ECT_REG_FMMU0       , 16 * 3    , &zbuf, EC_TIMEOUTRET3);  /* reset FMMU's */
    ecx_BWR(context->port, 0x0000, ECT_REG_SM0         , 8 * 4     , &zbuf, EC_TIMEOUTRET3);  /* reset SyncM */
-   b = 0x00; 
-   ecx_BWR(context->port, 0x0000, ECT_REG_DCSYNCACT   , sizeof(b) , &b, EC_TIMEOUTRET3);     /* reset activation register */ 
+   b = 0x00;
+   ecx_BWR(context->port, 0x0000, ECT_REG_DCSYNCACT   , sizeof(b) , &b, EC_TIMEOUTRET3);     /* reset activation register */
    ecx_BWR(context->port, 0x0000, ECT_REG_DCSYSTIME   , 4         , &zbuf, EC_TIMEOUTRET3);  /* reset system time+ofs */
    w = htoes(0x1000);
    ecx_BWR(context->port, 0x0000, ECT_REG_DCSPEEDCNT  , sizeof(w) , &w, EC_TIMEOUTRET3);     /* DC speedstart */
@@ -317,6 +317,7 @@ int ecx_config_init(ecx_contextt *context, uint8 usetable)
    uint8 SMc;
    uint32 eedat;
    int wkc, cindex, nSM;
+   uint16 val16;
 
    EC_PRINT("ec_config_init %d\n",usetable);
    ecx_init_context(context);
@@ -327,8 +328,8 @@ int ecx_config_init(ecx_contextt *context, uint8 usetable)
       for (slave = 1; slave <= *(context->slavecount); slave++)
       {
          ADPh = (uint16)(1 - slave);
-         context->slavelist[slave].Itype =
-            etohs(ecx_APRDw(context->port, ADPh, ECT_REG_PDICTL, EC_TIMEOUTRET3)); /* read interface type of slave */
+         val16 = ecx_APRDw(context->port, ADPh, ECT_REG_PDICTL, EC_TIMEOUTRET3); /* read interface type of slave */
+         context->slavelist[slave].Itype = etohs(val16);
          /* a node offset is used to improve readibility of network frames */
          /* this has no impact on the number of addressable slaves (auto wrap around) */
          ecx_APWRw(context->port, ADPh, ECT_REG_STADR, htoes(slave + EC_NODEOFFSET) , EC_TIMEOUTRET3); /* set node address of slave */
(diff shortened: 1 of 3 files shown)


soem/ethercatconfig.c
```diff
@@ -482,8 +483,8 @@ int ecx_config_init(ecx_contextt *context, uint8 usetable)
             context->slavelist[slave].SM[1].StartAddr = htoes(context->slavelist[slave].mbx_ro);
             context->slavelist[slave].SM[1].SMlength = htoes(context->slavelist[slave].mbx_rl);
             context->slavelist[slave].SM[1].SMflags = htoel(EC_DEFAULTMBXSM1);
-            context->slavelist[slave].mbx_proto =
-               ecx_readeeprom2(context, slave, EC_TIMEOUTEEP);
+            val = ecx_readeeprom2(context, slave, EC_TIMEOUTEEP);
```
nakarlsson: skip adding uint32_t val, use the same look'n'feel as done earlier in the file, for example L388 "eedat = etohl(ecx_readeeprom2(context, slave, EC_TIMEOUTEEP)); /* read mailbox offset */"

soem/ethercattype.h
```diff
@@ -526,6 +525,8 @@ typedef struct
   #define etohll(A) (A)
 
 #elif !defined(EC_LITTLE_ENDIAN) && defined(EC_BIG_ENDIAN)
+  #define put_unaligned32(val, ptr)        \
+    (memcpy((ptr), ((uint32*)&(val))+1, 4))
```
nakarlsson: Please explain how put_unaligned32 works for BE.

Why don't you need put put_unaligned64 fro BE

soem/ethercatconfig.c
Schlumpf: I used the temporary variable because `etohs()` is a macro. Fully expanded, it results in (on big-endian systems):
`((((uint32)(ecx_readeeprom2(context, slave, 20000)) & 0xff000000) >> 24) | \ (((uint32)(ecx_readeeprom2(context, slave, 20000)) & 0x00ff0000) >> 8) | \ (((uint32)(ecx_readeeprom2(context, slave, 20000)) & 0x0000ff00) << 8) | \ (((uint32)(ecx_readeeprom2(context, slave, 20000)) & 0x000000ff) << 24))`.
So `ecx_readeeprom2()` would called 4 times, instead of once.

nakarlsson: Ok, that's no good, that would need to be addressed.  A few of those and other ethercat commands but isolated to ethercatconfig.c, would you mind fixing it in this PR?

soem/ethercattype.h
Schlumpf: If I understood correctly, `put_unaligned32()` should copy the used 4 bytes of an 8 byte buffer variable to a destination buffer. It is used in `ecx_siigetbyte()`, only.

If you write an 32 bit value to an 64 bit variable, the highest 4 bytes are used:
Value: `0x12345678` -> Memory: `78 56 34 12 00 00 00 00`
On big-endian systems, value use the last 4 bytes:
Value: `0x12345678` -> Memory: `00 00 00 00 12 34 56 78`
If you only copy the first 4 bytes of the variable on big-endian systems, you would get 0x0 every time. Here you have to copy the last 4 bytes.

This change differentiates between the systems and copies the second 4 bytes on big-endian systems. I cannot use the  `put_unaligned64()` macro, because it copies 8 bytes instead of 4.

nakarlsson: The reason for using put_alignedX is that &(context->esibuf[eadr << 1]) might not be aligned, that might trigger exceptions on MCU arch that can't handle that.

So, if I understand you correct, the issue is only put_aligned32 due to the fact that edat is an uint64_t I don't think changing put_unaligned32 is the right way to go, would this work?

            put_unaligned32(edat, &(context->esibuf[eadr << 1]));
            edat = htoell(edat);
            cnt = 4;

Schlumpf: That wouldn't be correct. `htoell()` would swap all bytes, so the order of the needed 4 bytes would changed.
The problem is the use of a `uint64` as a buffer for a 32 bit value. In my new commit, I solved the problem otherwise by shifting the needed bits to the upper 4 bytes.

soem/ethercatconfig.c
Schlumpf: Done in new commit

soem/ethercattype.h
```diff
@@ -513,10 +513,10 @@ typedef struct
 #define put_unaligned64(val, ptr)        \
   (memcpy((ptr), &(val), 8))
 
-#if !defined(EC_BIG_ENDIAN) && defined(EC_LITTLE_ENDIAN)
-  #define put_unaligned32(val, ptr)        \
-    (memcpy((ptr), &(val), 4))
+#define put_unaligned32(val, ptr)        \
```
nakarlsson: Can you revert the changes in this file all together? Doesn't need to be part of the PR anymore.?

soem/ethercatmain.c
```diff
@@ -370,6 +370,10 @@ uint8 ecx_siigetbyte(ecx_contextt *context, uint16 slave, uint16 address)
          /* 4 byte response */
          else
          {
+#ifdef EC_BIG_ENDIAN
```
nakarlsson: I get you point but I'd like to make it more clear what's going on. How about making it more explicit?

   uint64 edat64;
   uint32 edat32;
...
         edat64 = ecx_readeepromFP (context, configadr, eadr, EC_TIMEOUTEEP);
         /* 8 byte response */
         if (context->slavelist[slave].eep_8byte)
         {
            put_unaligned64(edat64, &(context->esibuf[eadr << 1]));
            cnt = 8;
         }
         /* 4 byte response */
         else
         {
            edat32 = (uint32)edat64;
            put_unaligned32(edat32, &(context->esibuf[eadr << 1]));
            cnt = 4;
         }

nakarlsson:
2 minor things
1. Remove ethercattype.h changes from the PR, delete first commit or sqush them?
2. Try to make the uint64 -> uint32 conversion more obvious, I made an inline comment with proposal.

Otherwise it looks OK and works as expected on Little Endian.

soem/ethercatmain.c
Schlumpf: You are right, will do

Schlumpf:
Done :)

soem/ethercattype.h
Schlumpf: Done

nakarlsson:
thanks, I'll close #174 for now, if more issues are left please write a new issue

PR #230, opened 2018-11-18
update PO2SOconfig to add a ecx_contextt* in callback argument.

fred-lee-sv:
Update PO2SOconfig callback to add ecx_contextt* context as a parameter in callback function in order to make it convenient for using the interface 'ecx_xxx' (not using the EC_VER1 apis)


nakarlsson:
it makes sense, the only thing is the API change, historically we've tried to have a stable API hence the VER1. In this particular case it is hard to do any other way if we want this functionality,
@ArthurKetels , your opinion?

ArthurKetels:
The stable API is very important. So I would go for a 2nd hook and just leave the old one as it is. It is not performance critical.  There are some other spots with similar problems. Marking the old hook as deprecated is fine with me.

PR #240, opened 2018-12-22
WIP: Add Mac support

tjacobs:


ArthurKetels:
As far a I know you need to use libpcap to be able to send raw packets. The standard socket interface will not allow non IP packets to be send. So I guess it is easier to look at the windows implementation of nicdrv.c as far raw packet handling is concerned.

nakarlsson:
Are there any work ongoing on this PR?

nakarlsson:
done via https://github.com/OpenEtherCATsociety/SOEM/pull/286

PR #243, opened 2019-01-11
Bug with first slave with many PDOs and DC enabled. 

lucap85:
If the first slave has DC enabled and enouh PDOs to need more than [EC_MAXLRWDATA - EC_FIRSTDCDATAGRAM] bytes in the process image, the first segment's size will be zero instead of EC_FIRSTDCDATAGRAM and this cause the wrongly reading for the wkc at receiving time.

### DIFF
--- a/soem/ethercatconfig.c
+++ b/soem/ethercatconfig.c
@@ -1171,7 +1171,7 @@ int ecx_config_map_group(ecx_contextt *context, void *pIOmap, uint8 group)
                oLogAddr = LogAddr;
                if ((segmentsize + diff) > (EC_MAXLRWDATA - EC_FIRSTDCDATAGRAM))
                {
-                  context->grouplist[group].IOsegment[currentsegment] = segmentsize;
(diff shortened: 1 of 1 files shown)


nakarlsson:
I've like to get some additional information to verify the fix
Could you add a slaveinfo dump with that 1 slave
Can capture a wireshark from simple_test with that 1 slave

antol85:
Hi,

here they are:

[slaveinfo_and_simpletest_results.zip](https://github.com/OpenEtherCATsociety/SOEM/files/2835817/slaveinfo_and_simpletest_results.zip)

ArthurKetels:
Unfortunately the wireshark dump does not show how the fix should work. Slave 1 has an input size smaller than [EC_MAXLRWDATA - EC_FIRSTDCDATAGRAM] (1486 - 20). So everything works normal.

Also I think the fix is incorrect. When having a first slave with a very large PDO size then the correct solution would be to remove the LRW datagram when the segment size = 0 (or replace it with a NOP datagram).

nakarlsson:
This fix needs some re-work , please update and repost a proper fix as Arthur suggested

hefloryd:
Fixed by #907

PR #249, opened 2019-02-02
Re-init when error acknowledge fails

MCFurry:
I found that some (old) slaves do not respond to EC_STATE_SAFE_OP + EC_STATE_ACK correctly. The work-counter is correctly updated but the state is stuck in EC_STATE_SAFE_OP + EC_STATE_ERROR. If this acknowledge fails multiple times I suggest to do a re-init of that slave. This succeeds in bringing the slave back to EC_STATE_OPERATIONAL.

The fix works, but please comment whether this is the neatest implementation.

### DIFF
--- a/soem/ethercatmain.h
+++ b/soem/ethercatmain.h
@@ -229,6 +229,8 @@ typedef struct ec_slave
    int              (*PO2SOconfig)(uint16 slave);
    /** readable name */
    char             name[EC_MAXNAME + 1];
+   /** To keep track of failed SAFE_OP -> OP attempts **/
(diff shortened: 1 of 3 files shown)


nakarlsson:
I don't think this has to/should invlove generic SOEM, eg most of the changes are in the EtherCAT Master Application, one variable is stored but not used by the SOEM lib. It would be different if it concern an ESC HW.

What actual slaves are we talking about , vendor and product ?? : ??

nakarlsson:
Seems not active, please repost or open an issue for discussion first

PR #252, opened 2019-02-05
Patchset

claudioscordino:
This patchset:
 - fixes several typos in comments across all project;
 - makes a couple of additions to README and .gitignore;
 - isolates multi-thread code;
 - introduces platform-specific EC_PRINT.

The last two items are preparatory for supporting a non-POSIX RTOS like ERIKA Enterprise (on-going porting).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ install
 *~
 /doc/latex
 /doc/html
+tags
--- a/README.md
+++ b/README.md
@@ -29,6 +29,12 @@ Linux
    * `cmake ..`
    * `make`
 
+Documentation
+-------------
+
+See https://openethercatsociety.github.io/doc/soem/
+
+
 Want to contribute to SOEM or SOES?
 -----------------------------------
 
--- a/doc/soem.dox
+++ b/doc/soem.dox
@@ -10,7 +10,7 @@
  * how an EtherCAT master functions and how it interacts with EtherCAT slaves.
  *
  * As all applications are different SOEM tries to not impose any design architecture.
- * Under Linux it can be used in generic user mode, PREEMPT_RT or Xenomai. under Windows
+ * Under Linux it can be used in generic user mode, PREEMPT_RT or Xenomai. Under Windows
  * it can be used as user mode program.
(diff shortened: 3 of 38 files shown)


ArthurKetels:
Good patch set. Nice to see another target being supported. For me ok to merge.

claudioscordino:
> The rationale given for [fa0606a](https://github.com/OpenEtherCATsociety/SOEM/commit/fa0606a36c16548c42f3a58e2d74560aaba67dd5) is a little confusing - OSAL_THREAD is an abstraction layer for any threading framework, not just pthreads. The patch makes sense though, in order to allow running SOEM on bare metal for instance.

Hi Hans. Do you want me to re-submit the patchset with a different commit message, or you can change such message when pulling ?
Many thanks and best regards.

ArthurKetels:
Hi Claudio, do you have any timescale when you want to merge the full target support? 
Also it would be easier for me if you could change the commit message.

claudioscordino:
Hi Arthur.
No problem for changing the commit message. Do you mind sending me a sentence that you feel appropriate ?
Concerning the full target support, indeed I already finished the porting and tested with a Beckhoff slave. My original plan was to merge all the stuff in two steps: the preparatory patches that I've already sent, then wait another month for further testing. However, if you prefer, we can do it with just one pull request. Let me know how you prefer to proceed. Many thanks.

nakarlsson:
FYI a CLA has been signed

ArthurKetels:
Hello Claudio,

Below text suggestion would be more appropriate.
"Avoid OSAL_THREAD usage when EC_MAX_MAPT=1. Rationale : Allows SOEM to be used in a OS that does not support threads."

I agree it would be best to split the commits and let the target support soak a little while before merging. It was just out of curiosity. I would like to have a try on ERIKA and jailhouse to see if we could boost the performance of EtherCAT on COTS hardware. For real high performance systems Linux alone just doesn't cut it.

claudioscordino:
Ok. I'm going to kill this pull request and start a new one with the right commit message. Thanks.

PR #260, opened 2019-03-08
improve groups, make memory usage more efficient, fix erroneous byte …

nakarlsson:
…calculations and dc frame handling

Use case, separate IOmap per group, it is undocumented if/how usage of _map_groups( .., >0) should/will work. 

This PR will handle these scenarios where the second would be legacy

char iomap1[4096];
char iomap2[2048];
char iomap3[1024];

ecx_config_map_group(iomap1, 1);
ecx_config_map_group(iomap2, 2);
ecx_config_map_group(iomap3, 3);

alt 

char iomap[4096];  
ecx_config_map(iomap);  typ ecx_config_map_group(iomap, 0);

### DIFF
--- a/soem/ethercattype.h
+++ b/soem/ethercattype.h
@@ -71,6 +71,8 @@ extern "C"
 #define EC_MAXEEPBUF       EC_MAXEEPBITMAP << 5
 /** default number of retries if wkc <= 0 */
 #define EC_DEFAULTRETRIES  3
+/** default group size in 2^x */
+#define EC_LOGGROUPOFFSET 16
 
 /** definition for frame buffers */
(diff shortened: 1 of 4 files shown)


soem/ethercattype.h
```diff
@@ -71,6 +71,8 @@ extern "C"
 #define EC_MAXEEPBUF       EC_MAXEEPBITMAP << 5
 /** default number of retries if wkc <= 0 */
 #define EC_DEFAULTRETRIES  3
+/** default group size */
```
ArthurKetels: /** default group size in 2^x */

nakarlsson:
I'be push'ed an update with the comment fix.
I've done some basic testing with multiple groups, verifed config and IOmap

PR #271, opened 2019-03-27
Tutorial: explain how to access CoE SDOs and PDOs

claudioscordino:
Trivial patch reporting some useful information already provided as
GitHub issues.


claudioscordino:
Feel free to rephrase or modify my commit.
The important thing IMO is that the tutorial _at least_ mentions PDOs and SDOs and how to access them using an alternative method than IOmap.

PR #275, opened 2019-04-01
Remove possible racecondition of set buf EMPTY

nakarlsson:
Remove "extra" clear buffer on lost frame in the nic driver, it is done unconditional by the thread/user who allocated the buffer.

I've run win32  simpletest sync processdata and async sdo reads


ArthurKetels:
Thanks for catching that! It is not only "possible" but a very real race condition. This is the first good explanation why the multi-threaded config_map fails sometimes.

PR #276, opened 2019-04-01
Add ERIKA Enterprise RTOS support

claudioscordino:
Signed-off-by: Claudio Scordino <claudio@evidence.eu.com>
Signed-off-by: Luca Cuomo <l.cuomo@evidence.eu.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,11 @@ Linux
    * `cmake ..`
    * `make`
 
+ERIKA Enterprise RTOS
+---------------------
+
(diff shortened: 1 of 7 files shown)


claudioscordino:
Yes, sorry, it's a left over from testing. I will create a new pull request.

Concerning the possibility of testing the new support: we currently support SOEM only on x86-64 platforms with Intel PCIe i210 Ethernet card and Xen hypervisor. If you have such a platform, we can help you in setting up a simple demo. However, I wonder if the hardware platform is too specific for upstreaming the code to the SOEM mainline tree... what is the policy for including new supports ? Thanks.

PR #280, opened 2019-04-08
Fix to enable building the SOEM library VER 2 with -werror

nakarlsson:
Handle unsued variables


nakarlsson:
Note, this only fix building the library. Building the test folder still fail due to applications being implemented VER 1. That is another topic / no-issue

Issue #53, opened 2016-08-18
osal.c: absolute vs. relative time (Windows)


Apollo3zehn:
On Windows systems, the function **ecx_configdc** uses **osal_current_time()** to initialize the mastertime variable:

```
mastertime = osal_current_time();
```

The function **osal_current_time** calls **osal_gettimeofday** which relies on **QueryPerformanceFrequency** and **QueryPerformanceCounter**. When precise but relative time is needed (e.g. for timers), **QueryPerformanceCounter** is a good choice. But for initialization of DC, an absolute time would be more meaningful. 

[Here](https://msdn.microsoft.com/en-us/library/windows/desktop/dn553408%28v=vs.85%29.aspx) is a guidance on how to get time stamps in windows. In section **Guidance for acquiring time stamps** you find the following statements:

> When you need time stamps with a **resolution of 1 microsecond** or better and you don't need the time stamps to be synchronized to an external time reference, choose **QueryPerformanceCounter**, KeQueryPerformanceCounter, or KeQueryInterruptTimePrecise.
> 
> [...]
> 
> **If resolution of 10 to 16 milliseconds is sufficient**, you can use GetTickCount64, QueryInterruptTime, QueryUnbiasedInterruptTime, KeQueryInterruptTime, or KeQueryUnbiasedInterruptTime to obtain time stamps that aren't synchronized to an external time reference. For **UTC-synchronized** time stamps, use **GetSystemTimeAsFileTime** or KeQuerySystemTime.

The Ke*-functions are for kernel mode, which is not applicable here.

So, for versions <= Windows 7, there is a trade-off between precision and synchronization:

**QueryPerformanceCounter**: precise, not synchronized (relative time since windows boot)
**GetSystemTimeAsFileTime**: unprecise, UTC synchronized

So I would propose to use **GetSystemTimeAsFileTime** for DC initialization and **GetSystemTimeAsPreciseFileTime** when version >= Windows 8 because it combines both advantages.
###### 

In windows 7, I use the following function:

```
int osal_getabsolutetimeofday(struct timeval *tv, struct timezone *tz)
{
    FILETIME SystemTime;
    int64 SystemTime64, usecs;
    int64 Offset = -134774LL * 86400LL * 1000LL * 1000LL * 10LL; // Difference between 1601 and 1970 in 100 ns.

    timeBeginPeriod(1);
    GetSystemTimeAsFileTime(&SystemTime);
    timeEndPeriod(1);

[...]

PR #283, opened 2019-04-19
osal.c (Windows): Use absolute time instead of relative time.

Apollo3zehn:
Based on issue #53.

Currently - on Windows - the function `osal_gettimeofday` provides the number of ticks since system boot, so it is not the _time of day_. To initialize DC correctly and also to provide meaningful time values in all the error functions (like `ecx_mbxerror` and `ecx_SoEerror`), I would like to add extended functionality to `osal.c`:

For this PR I renamed `osal_gettimeofday` to `osal_getrelativetime` and added a new implementation for `osal_gettimeofday` to now provide the absolute time. It is not advisable to use the absolute time everywhere, because `GetSystemTimeAsFileTime` is slower than `QueryPerformanceCounter`. For relative measurements, `QueryPerformanceCounter` is still a good choice.

I did not include `GetSystemTimePreciseAsFileTime`, which delivers a much more precise time, because Windows 7 and lower do not support this API and I don't know how to make the calls between `GetSystemTimePreciseAsFileTime` and `GetSystemTimeAsFileTime` conditionally, depending on the OS version.

### DIFF
--- a/osal/win32/osal.c
+++ b/osal/win32/osal.c
@@ -12,7 +12,7 @@ static double qpc2usec;
 
 #define USECS_PER_SEC     1000000
 
-int osal_gettimeofday (struct timeval *tv, struct timezone *tz)
+int osal_getrelativetime(struct timeval *tv, struct timezone *tz)
 {
    int64_t wintime, usecs;
    if(!sysfrequency)
@@ -23,6 +23,33 @@ int osal_gettimeofday (struct timeval *tv, struct timezone *tz)
    }
    QueryPerformanceCounter((LARGE_INTEGER *)&wintime);
    usecs = (int64_t)((double)wintime * qpc2usec);
+   tv->tv_sec = (long)(usecs / 1000000);
+   tv->tv_usec = (long)(usecs - (tv->tv_sec * 1000000));
+
+   return 1;
+}
+
+int osal_gettimeofday(struct timeval *tv, struct timezone *tz)
+{
+   FILETIME system_time;
+   int64 system_time64, usecs;
+
+   /* The offset variable is required to switch from Windows epoch (January 1, 1601) to
+    * Unix epoch (January 1, 1970). Number of days between both epochs: 134.774
+    *
+    * The time returned by GetSystemTimeAsFileTime() changes in 100 ns steps, so the
+    * following factors are required for the conversion from days to 100 ns steps:
+    *
+    * 86.400 seconds per day; 1.000.000 microseconds per second; 10 * 100 ns per microsecond
+   */
+   int64 offset = -134774LL * 86400LL * 1000000LL * 10LL;
+
+   GetSystemTimeAsFileTime(&system_time);
+
+   system_time64 = ((int64)(system_time.dwHighDateTime) << 32) + (int64)system_time.dwLowDateTime;
+   system_time64 += offset;
+   usecs = system_time64 / 10;
+
    tv->tv_sec = (long)(usecs / 1000000);
    tv->tv_usec = (long)(usecs - (tv->tv_sec * 1000000));
 
@@ -58,7 +85,7 @@ void osal_timer_start (osal_timert *self, uint32 timeout_usec)
    struct timeval timeout;
    struct timeval stop_time;
 
-   osal_gettimeofday (&start_time, 0);
+   osal_getrelativetime (&start_time, 0);
    timeout.tv_sec = timeout_usec / USECS_PER_SEC;
    timeout.tv_usec = timeout_usec % USECS_PER_SEC;
    timeradd (&start_time, &timeout, &stop_time);
@@ -73,7 +100,7 @@ boolean osal_timer_is_expired (osal_timert *self)
    struct timeval stop_time;
    int is_not_yet_expired;
 
-   osal_gettimeofday (&current_time, 0);
+   osal_getrelativetime (&current_time, 0);
    stop_time.tv_sec = self->stop_time.sec;
    stop_time.tv_usec = self->stop_time.usec;
    is_not_yet_expired = timercmp (&current_time, &stop_time, <);


nakarlsson:
I get the point about relative and absolute time, but what is the purpose of this PR? SOEM still use relative time internally as it should since performance is important. Are you planning to provide a second PR that introduce usage of absolute time usage internally when reporting errors?

If not I can't see the purpose since the user call would end up in the application that I persume don't use SOEM osal gettimeofday anyway.

Apollo3zehn:
I initially thought absolute time is used everywhere, except Windows. Therefore, for this answer, I went through all ```osal.c``` implementations and found that ```osal_gettimeofday()``` is implemented incosistently accross list of supported OS. The following is an overview of the implementations: 

- [intime](https://github.com/OpenEtherCATsociety/SOEM/blob/master/osal/intime/osal.c#L25) - absolute time
- [linux](https://github.com/OpenEtherCATsociety/SOEM/blob/master/osal/linux/osal.c#L34) - relative time
- [rtems](https://github.com/OpenEtherCATsociety/SOEM/blob/master/osal/rtems/osal.c#L34) - relative time
- [erika](https://github.com/OpenEtherCATsociety/SOEM/blob/master/osal/erika/osal.c#L30-L33) - unclear (does ```osEE_x86_64_tsc_read()``` return absolute time or is it a simple counter in the processor?). If it is relative, why is the epoch added?
- [rtk](https://github.com/OpenEtherCATsociety/SOEM/blob/master/osal/rtk/osal.c#L52) - relative time
- [vxworks](https://github.com/OpenEtherCATsociety/SOEM/blob/master/osal/vxworks/osal.c#L87) - absolute time
- [win32](https://github.com/OpenEtherCATsociety/SOEM/blob/master/osal/win32/osal.c#L24) - relative time

As mentioned, I was expecting from the method name '```gettimeofday()```' that absolute time is used on all OS except Windows. Based on this, the purpose of this PR was to unify the implementations. But it seems that this is not easily accomplished due to the possiblity of live locks, if I get it right from the comment in the linux implementation.

In my opinion, using absolute time (or *real time*) make sense in two places: For DC initalization and for error messages. This PR provides both:

- [DC](https://github.com/OpenEtherCATsociety/SOEM/blob/master/soem/ethercatdc.c#L271) uses ```osal_current_time()``` which in turn calls into the new implementation of ```osal_gettimeofday()```
- [ecx_SoEerror](https://github.com/OpenEtherCATsociety/SOEM/blob/8832ef0f2fa09fb0b06acaca239b51263ca1878c/soem/ethercatsoe.c#L52) and other error functions also call ```osal_current_time()``` to get the time.

With this PR the call hierachy becomes:

**absolute:**
```ecx_SoEerror()``` -> ```osal_current_time()``` -> ```osal_gettimeofday()```
```ecx_mbxerror()``` -> ```osal_current_time()``` -> ```osal_gettimeofday()```
```ecx_SDOerror()``` -> ```osal_current_time()``` -> ```osal_gettimeofday()```

```ecx_configdc()``` -> ```osal_current_time()``` -> ```osal_gettimeofday()``` 

**relative:**
```<time critical method>``` -> ```osal_timer_start``` -> ```osal_getrelativetime()```
```<time critical method>``` -> ```osal_timer_is_expired```  -> ```osal_getrelativetime()```

You can see from the overview that the time critical methods (```osal_timer_start()``` and ```osal_timer_is_expired()```) now call into ```osal_getrelativetime()```. **So there is no impact in time critical sections.**

To answer your questions more compact:

#### What is the purpose of this PR?
- My personal motivation behind this is to use DC time as source for measurement data timestamps (see also #287). For this, absolute time is useful. 
- SOEM errors should provide absolute time.
- I think a method called 'gettimeofday' should return the time of the day (instead of a random number).
- The document [Hardware Data Sheet Section I - Technology](https://download.beckhoff.com/download/document/io/ethercat-development-products/ethercat_esc_datasheet_sec1_technology_2i2.pdf) provides a definition of the system time (e.g. epoch start at 01.01.2000) and I think it is advantageous if all master implementations align with this definition.

#### Are you planning to provide a second PR that introduce usage of absolute time usage internally when reporting errors?
- In this PR, the above listed error functions already get the absolute time.

I do understand if you think this PR will introduce new inconsistencies with other OS's since they use either absolute **or** relative time. And with this PR only Windows uses both variants. But long term, I think the handling with absolute time should be unified on all OS implementations.

nakarlsson:
Thanks for the analysis, it makes sense and I think aligning Win32 with Linux(eg PCs) when it comes to functionallity is a good thing. @ArthurKetels get final call since he did the Win32 port.
I mainly work with embedded systems and quite often don't have absolute time at all.

Please squash to a single commit.

ArthurKetels:
I am happy to take the PR. And also from me, thanks for the analysis. I am fully aware of the unaligned implementations of time for the targets we support. SOEM itself is happy with any version of local time, as long it is fast and monotonous. The application however might be more interested in absolute time. 

But then I must complain about the utter garbage all OS have for absolute time. They are at best usable for file time stamps. But have nothing to do with control quality time keeping. 

For accurate measurements it is just as important to have precise time as having precise measurement values. A measurement is a vector, a measured quantity of interest at a moment in time. (and yes, this implies a clock as a quantity of interest).

Because our electronic hardware has no concept of time other than an arbitrary clock don't fool yourself you could have something like absolute time. It will only be a rough approximation.  Depending on your use case it could be acceptable but never the less an approximation. Other use cases might require much more accurate approximations. 

So in a way, all attempts to make an absolute time are still relative times. Then just as well stick with simple relative times and have some good estimator functions that can generate an acceptable guess  to what absolute time belongs a specific relative time stamp.

As an example a struct I use for accurate timing:

struct {
uint64  timestamp;
int  estimate_id;
uint64  estimate_sec;
uint64  estimate_femtosec;
double drift;
double drift_derrivative;
} timevector_t;

The estimate is to a reference clock with estimate_id. This also allows to transpose the timestamp from one reference clock to another. And at the same time keeping the original timestamp.

--end rant mode --

Apollo3zehn:
Thanks @ArthurKetels for the explanation!

PR #292, opened 2019-05-08
Add possibiliy to set endian target by hardware layer

Schlumpf:
Now it is possible to set a EC_LITTLE_ENDIAN or EC_BIG_ENDIAN in the
osal_defs.h file or by compile option. If no endian is defined, the default
EC_LITTLE_ENDIAN is used.


nakarlsson:
thanks, make sense

PR #340, opened 2019-09-23
 Update to CMakeLists.txt

giellamoswhard:
Hi Guys, 
I've somewhat modernized the CMakeLists.txt so it can be used with find_package.
I don't know if you are interested (and by the way I only tested this in a linux environment).

Bye.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.4)
+cmake_minimum_required(VERSION 2.8.12)
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/Modules")
 project(SOEM C)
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -60,21 +60,38 @@ file(GLOB SOEM_HEADERS soem/*.h)
...
-
 add_library(soem STATIC
   ${SOEM_SOURCES}
   ${OSAL_SOURCES}
   ${OSHW_SOURCES}
   ${OSHW_EXTRA_SOURCES})
 target_link_libraries(soem ${OS_LIBS})
 
+target_include_directories(soem PUBLIC 
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}//soem>
```
hefloryd: Are the double forward-slashes needed?

giellamoswhard: Hi, no they are not.

hefloryd:
Thanks, the PR has been squashed and merged.

PR #360, opened 2019-12-18
Add cmake option to enable PIC

mockbutler:
Allows library to be used in the creation of dynamic libraries.
Not enabled by default.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,8 @@ set(SOEM_INCLUDE_INSTALL_DIR include/soem)
 set(SOEM_LIB_INSTALL_DIR lib)
 set(BUILD_TESTS TRUE)
 
+option(SOEM_PIC "Compile library as PIC, suitable for linking into a dynamic library." FALSE)
+
 if(WIN32)
   set(OS "win32")
   include_directories(oshw/win32/wpcap/Include)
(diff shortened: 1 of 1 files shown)


nakarlsson:
ping @hefloryd

hefloryd:
-fPIC is added automatically by cmake if BUILD_SHARED_LIBS is enabled. Currently of course a static library is always built, so you will also have to remove STATIC from the cmakelist, but I guess you would need to do that anyway in order to build a shared library. 

Note that PR #343 modernizes the cmakelist and lets you build a shared library without editing the cmakelists (just by setting BUILD_SHARED_LIBS). Please try it out if you have the time.

mockbutler:
@hefloryd For my purposes I want to build the static library with -fPIC. This library is then linked into another project which produces a dynamic library. The second project provides wrapper code for a managed language. So linking against an -fPIC static library simplifies the distribution to a single native library, precluding dynamic library version mismatch.

This use case is atypical, but I didn't think unique. So I thought might be useful to support this almost hybrid target. There are other ways to achieve this, but using CMake allows the broad platform support.

lepalom:
I agree with @mockbutler and I would like to add that it's a typical case to use SOEM and build a shared library to provide a wrapper for a managed language. 

Also, the BUILD_SHARED_LIBS  is referred to build a shared lib when you have add_library, so it's not clear why to set this parameter to add the -PIC flag.

PR #382, opened 2020-03-17
Added example with DC-sync

stex2005:
I reviewed the redundancy test by Arthur and added a working example to use SOEM with DC-sync.
Tested with an EPOS4 EtherCAT slave. Unfortunately, I got lot of disconnections at 1kHz, but still it works.


CMakeLists.txt
```diff
@@ -100,4 +100,5 @@ if(BUILD_TESTS)
   add_subdirectory(test/linux/slaveinfo)
   add_subdirectory(test/linux/eepromtool)
   add_subdirectory(test/linux/simple_test)
+  add_subdirectory(test/linux/dcsync_test)
```
stex2005: This can be deleted to pass tests

PR #393, opened 2020-04-18
As root message updated in simple_test

DeadZen:
Happened to spot this, just corrected Execute


nakarlsson:
Thanks for improving SOEM

Issue #392, opened 2020-04-17
MacOS Catalina Slave Power Cycle Problem


berkaydeniz:
Every time I power cycle the slave the assertion on line 443 of nicdrv.c is triggered
```
Assertion failed: (0), function ecx_inframe, file <---this part is removed for privacy reasons--->oshw/macosx/nicdrv.c, line 443.
```
and thus the software crashes. If I comment out the line, `//assert(0);` things work just fine. Obviously this assertion is in place for a purpose. Are there any general rules for handling a slave power cycle that I am missing or is this a bug?

Currently using commit version `7fe05b5b`.

Note that the same software works perfectly fine with no crashes when built under Ubuntu. So this is specifically triggered on MacOS Catalina.

ArthurKetels:
Looks like a frame is duplicated by the OS network stack. Can you make a wireshark trace?

nakarlsson:
It think the assert can be removed. The frame have been abandoned.

Why it happens is strange, specially in a single threaded SOEM application. Best guess, the frame get buffered in some send Q when the power cycle occur. During the power cycle SOEM try to receive it and fails, marking the buffer EMPTY.
But again, think the assert is to drastic, linux don’t assert. We don’t assert in other plattforms either.

nakarlsson:
Fixed in #399

PR #399, opened 2020-05-07
assert is removed from macos version of nicdrv.c

berkaydeniz:
As discussed in [Issue #392](https://github.com/OpenEtherCATsociety/SOEM/issues/392), this assert only exists for macos and is an overkill. Removing it should not affect anything. On the contrary, for my use case, it eliminates a huge headache.


nakarlsson:
Thanks for improving SOEM

Fix #392

PR #431, opened 2020-08-14
EoE and Distributed clock fixes

### LINKED ISSUE #421: EoE potentially invalid check data_offset > eoedatasize in ethercateoe.c
I have been experimenting with the tests/linux/eoe_test/eoe_test.c file and found that to make the GetIP work I had to comment out the check 

> if(data_offset > eoedatasize)

around line 300

The SSC 5.12 Doesn't seem to expect this to match.

Note that there is also a bug in the SSC re reading the IP address from aIpAdd in the event where MBX_16BIT_ACCESS is not set, see post in ETG Dev Forum.

### LINKED ISSUE #432: Wrong unit for master time calculation in ecthercatdc
Found and discussed in https://github.com/OpenEtherCATsociety/SOEM/issues/345#issuecomment-542614878 and https://github.com/OpenEtherCATsociety/SOEM/issues/345#issuecomment-625033118

nakarlsson:
EoE fix
Obvious fix, the optional IP parameters length should
only be included in total length if they are included.
fixes #421

DC fix
The correct factor for seconds should be 1000000000
fixes #432

### DIFF
--- a/soem/ethercateoe.c
+++ b/soem/ethercateoe.c
@@ -94,45 +94,43 @@ int ecx_EOEsetIp(ecx_contextt *context, uint16 slave, uint8 port, eoe_param_t * 
       EOE_HDR_LAST_FRAGMENT);
    EOEp->frameinfo2 = 0;
   
-   /* The EoE frame will include "empty" IP/DNS entries, makes wireshark happy.
-    * Specification say they are optional, TwinCAT include empty entries.
-    */
    if (ipparam->mac_set)
    {
       flags |= EOE_PARAM_MAC_INCLUDE;
       memcpy(&EOEp->data[data_offset], ipparam->mac.addr, EOE_ETHADDR_LENGTH);
+      data_offset += EOE_ETHADDR_LENGTH;
    }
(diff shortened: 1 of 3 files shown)


soem/ethercateoe.c
```diff
@@ -94,45 +94,43 @@ int ecx_EOEsetIp(ecx_contextt *context, uint16 slave, uint8 port, eoe_param_t *
...
       flags |= EOE_PARAM_MAC_INCLUDE;
       memcpy(&EOEp->data[data_offset], ipparam->mac.addr, EOE_ETHADDR_LENGTH);
+      data_offset += EOE_ETHADDR_LENGTH;
    }
-   data_offset += EOE_ETHADDR_LENGTH;
    if (ipparam->ip_set)
    {
       flags |= EOE_PARAM_IP_INCLUDE;
       EOE_ip_uint32_to_byte(&ipparam->ip, &EOEp->data[data_offset]);
+      data_offset += 4;
```
ArthurKetels, line 107: Perhaps better to do (here and other places below):
data_offset += sizeof(uint32_t);

PR #56, opened 2016-08-26
some improvement on FoE (large mailboxes, busy answer, firmware_update)

paperwork:

PR #483, opened 2021-02-21
Fix SyncManager configuration in firmware update example

### LINKED ISSUE #396: `firm_update.c` doesn’t disable the SyncManagers before reconfiguring them
I'm using the Beckhoff Slave Stack Code which initializes and enables the SyncManagers at bootup. `firm_update.c` tries to reconfigure SyncManager 1 to a different start address, but fails in doing so, because the SyncManager is already enabled. It needs to be disabled first, otherwise the new configuration is simply ignored.

Pull request #56 solves this issue.

jdeitmerg:
Hi,
I rediscovered #396 when playing with the firmware update example:
SyncManagers 0 and 1 are not configured correctly if the bootstrap mailbox parameters in EEPROM differ from the non-bootstrap ones.
The reason is that the SyncManagers are configured and enabled in ec_config_init(). This happens in accordance with the non-bootstrap mailbox settings in EEPROM. When the (differing!) bootstrap configuration is written later, the new *Physical Start Address*, *Length* and *Control* values are ignored as those can only be changed when the SyncManager is disabled.
The supplied fix disables the SyncManagers before reconfiguring them.
Just found out a similar fix is part of #56 
Fixes #396


test/linux/firm_update/firm_update.c
```diff
@@ -93,6 +93,14 @@ void boottest(char *ifname, uint16 slave, char *filename)
 			    (int)ec_slave[slave].SM[0].SMflags);
 			printf(" SM1 A:%4.4x L:%4d F:%8.8x\n", ec_slave[slave].SM[1].StartAddr, ec_slave[slave].SM[1].SMlength,
 			    (int)ec_slave[slave].SM[1].SMflags);
+
+			/* disable SM1 + SM2 so their offset + length can be changed from
```
nakarlsson, line 97: disable SM0 + SM1

test/linux/firm_update/firm_update.c
```diff
@@ -93,6 +93,14 @@ void boottest(char *ifname, uint16 slave, char *filename)
...
+			 * the values set in ec_config_init()
+			 */
+			uint8_t activate = 0;
+			ec_FPWR(ec_slave[slave].configadr, ECT_REG_SM0+6, 1, &activate, EC_TIMEOUTRET);
```
nakarlsson, line 101: * Make use of  ECT_REG_SM1ACT      

* Add ECT_REG_SM0ACT to ethercattype.h

nakarlsson:
@jdeitmerg check the comments

nakarlsson:
@jdeitmerg , can you squash the commits?

jdeitmerg:
Hi, thanks for the pointers. The implementation is better this way.

What do you think about making the firmware update example compile by default? That way it would be checked by the automatic build workflow. Shall I open another PR?

nakarlsson:
Did you investigate why the SM is kept enabled when going from PREOP->INIT before entering BOOT?
That is not in accordance to the standard, hence it could be worth mentioning why this fix is needed, and what it fixes.

nakarlsson:
@jdeitmerg , any motivation for adding this fix? Since, it is not according to standard we'd like some note on why it is needed.

regards

jdeitmerg:
I'm trying to figure out why things don't behave according to the standard. I'll let you know as soon as I know what's going on.

jdeitmerg:
Still haven't gotten around to finding the root cause of this. I did build a slave with the Beckhoff stack though, which passes the Conformance Test Tool. With that the firm_update.c example works correctly. Closing this because it seems to fix a symptom, not an actual bug.

PR #486, opened 2021-02-25
simple_test revamp

ntd:
The `simple_test` program has been rewritten from scratch with the
following major changes:

* portable codebase, so no splitting between different OS;
* no threads: they are not needed for such simple task;
* use newer APIs (i.e., ecx_... functions).

### DIFF
--- a/test/simple_ng/simple_ng.c
+++ b/test/simple_ng/simple_ng.c
@@ -1,0 +1,316 @@
+/** \file
+ * \brief Example code for Simple Open EtherCAT master
+ *
+ * Usage: simple_ng IFNAME1
+ * IFNAME1 is the NIC interface name, e.g. 'eth0'
+ *
+ * This is a minimal test.
+ */
+
+#include "ethercat.h"
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+
+typedef struct {
+    ecx_contextt    context;
+    char *          iface;
+    uint8           group;
+    int             roundtrip_time;
+
+    /* Used by the context */
+    uint8           map[4096];
+    ecx_portt       port;
+    ec_slavet       slavelist[EC_MAXSLAVE];
+    int             slavecount;
+    ec_groupt       grouplist[EC_MAXGROUP];
+    uint8           esibuf[EC_MAXEEPBUF];
+    uint32          esimap[EC_MAXEEPBITMAP];
+    ec_eringt       elist;
+    ec_idxstackT    idxstack;
+    boolean         ecaterror;
+    int64           DCtime;
+    ec_SMcommtypet  SMcommtype[EC_MAX_MAPT];
+    ec_PDOassignt   PDOassign[EC_MAX_MAPT];
+    ec_PDOdesct     PDOdesc[EC_MAX_MAPT];
+    ec_eepromSMt    eepSM;
+    ec_eepromFMMUt  eepFMMU;
+} Fieldbus;
+
+
+static void
+fieldbus_initialize(Fieldbus *fieldbus, char *iface)
+{
+    ecx_contextt *context;
+
+    /* Let's start by 0-filling `fieldbus` to avoid surprises */
+    memset(fieldbus, 0, sizeof(*fieldbus));
+
+    fieldbus->iface = iface;
+    fieldbus->group = 0;
+    fieldbus->roundtrip_time = 0;
+    fieldbus->ecaterror = FALSE;
+
+    /* Initialize the ecx_contextt data structure */
+    context = &fieldbus->context;
+    context->port = &fieldbus->port;
+    context->slavelist = fieldbus->slavelist;
+    context->slavecount = &fieldbus->slavecount;
+    context->maxslave = EC_MAXSLAVE;
+    context->grouplist = fieldbus->grouplist;
+    context->maxgroup = EC_MAXGROUP;
+    context->esibuf = fieldbus->esibuf;
+    context->esimap = fieldbus->esimap;
+    context->esislave = 0;
+    context->elist = &fieldbus->elist;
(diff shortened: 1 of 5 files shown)


ntd:
I had to force-push some commits to pass the windows build.

test/simple_test/simple_test.c
```diff
@@ -0,0 +1,320 @@
...
+
+    /* Used by the context */
+    uint8           map[4096];
+    ecx_portt       port;
+    ec_slavet       slavelist[EC_MAXSLAVE];
+    int             slavecount;
+    ec_groupt       grouplist[EC_MAXGROUP];
+    uint8           esibuf[EC_MAXEEPBUF];
+    uint32          esimap[EC_MAXEEPBITMAP];
+    /* I would say this needs to be elist[EC_MAXELIST],
```
nakarlsson, line 30: Why, a single ring of EC_MAXELIST error entries

nakarlsson:
thanks, it looks nice.
* How about keeping linux\simple_test asis , name the new simple_test something clever or some ref to ecx? ver2? What do you think.
* Can you squash the commits

ntd:
> * How about keeping linux\simple_test asis , name the new simple_test something clever or some ref to ecx? ver2? What do you think.

I was myself undecided if adding a new test or updating the old one. I'll rename it to `simple_ng`: if needed, `simple_test` can be dropped in a second time.

> * Can you squash the commits

Ok, I will.

test/simple_test/simple_test.c
ntd: I don't know what I'm talking about, but that the field name and the existence of `EC_MAXELIST` strongly suggested that `elist` should be an array (similar to e.g. `grouplist`) although it was not used in that way by the sample code.
  
A recent look at [`ec_eringt`](https://github.com/OpenEtherCATsociety/SOEM/blob/d548e9907db8ba2e4c0967f4769734255f97f371/soem/ethercatmain.h#L346-L352) shown me how this is really implemented, so this is just a case of unfortunate naming.

I will remove the comment.

test/simple_ng/simple_ng.c
```diff
@@ -0,0 +1,318 @@
...
+
+
+typedef struct {
+    ecx_contextt    context;
+    char *          iface;
+    uint8           group;
+    int             roundtrip_time;
+
+    /* Used by the context */
+    uint8           map[4096];
```
nakarlsson, line 23: Set the map to 0's? 

Or clear the entire filedbus map? Otherwiase you'll start with heap values, that can be confusing for users.      
memset(&fieldbus, 0, sizeof(fieldbus));

test/simple_ng/simple_ng.c
```diff
@@ -0,0 +1,318 @@
...
+{
+    ec_groupt *grp;
+    uint32 n;
+    int wkc, expected_wkc;
+
+    grp = fieldbus->grouplist + fieldbus->group;
+
+    wkc = fieldbus_roundtrip(fieldbus);
+    expected_wkc = grp->outputsWKC * 2 + grp->inputsWKC;
+    printf("%d usec  WKC %d", fieldbus->roundtrip_time, wkc);
```
nakarlsson, line 213: Add a fixed print format, otehrwise it becomes a big blur on the output.
Ex below is nicer to look at.
    printf("%3d usec  WKC %d", fieldbus->roundtrip_time, wkc);

nakarlsson:
I'm sorry for the delay, I've been really busy and when I've been free I haven't had a chance to test on a target.
Looks good , I have a few comments though.

test/simple_ng/simple_ng.c
ntd: This would limit the time to 1 ms. On some test I did in harsh conditions, I've seen roundtrip times **much** longer than that. For this reason I'd rather prefer to leave the length free.
Alternatively, considering the other fields are almost fixed-size, I can move it at the end of the line, so it can grow as it wishes without messing up everything.

test/simple_ng/simple_ng.c
ntd: Thank you for taking the time to review.
As I have no experience with SOEM, I strictly followed what already done in [ec_master.c](https://github.com/OpenEtherCATsociety/SOEM/blob/master/test/intime/ec_master/ec_master.c#L168), where `IOmap` is not initialized.
Not a big deal though, I will add the whole struct clearing as suggested: I'm already doing it for `ecx_contextt`, so it just needs to be extended to the whole `Fieldbus` struct.

ntd:
> Hi @ntd, I'm sorry for the delay

No problems, I didn't hold my breath :)

I just addressed the first comment but I have some reservation about the second one.

test/simple_ng/simple_ng.c
nakarlsson: Sure, as long as it is possible to see what’s is going on

test/simple_ng/simple_ng.c
nakarlsson: Sure , global variabels get 0’ed by default. Stack variabels doesn’t, unless they’re static

ntd: > Sure , global variabels get 0’ed by default.

You're right, never known about that.

test/simple_ng/simple_ng.c
ntd: Moving that field to the end was not feasible because while the roundtrip time is always present, I do not output some of the other fields if the wkc does not match. I just went to the padding root with more loose limits, i.e. 6 digits for the roundtrip time and 4 digits for the iteration counter.

ntd:
I just resolved the second comment: let me know if there is some other issue.

nakarlsson:
thanks

PR #495, opened 2021-04-01
Check al_state after request

alexkalmuk:
I faced a race condition using SOEM/SOES in a simulated environment (that is, without real hardware for slave). The following race condition occasionally happens in `test/linux/firm_update/firm_update.c` which leads to firm_update failure:

```cpp
	    if ( ec_config_init(FALSE) > 0 )
		{
			printf("%d slaves found and configured.\n",ec_slavecount);

			printf("Request init state for slave %d\n", slave);
			ec_slave[slave].state = EC_STATE_INIT;
			ec_writestate(slave);

			/* wait for slave to reach INIT state */
			ec_statecheck(slave, EC_STATE_INIT,  EC_TIMEOUTSTATE * 4);
```

During `ec_config_init` SOEM requests transition to PRE_OP state without waiting. There is a possible situation when we reach the next lines with the transition to INIT state before the previous transition was actually handled by the slave. I added several `ecx_statecheck` to fix this and making SOEM not relying on the "slave's speed".

P.S.
CMakeLists.txt is missing in firm_update/ (I created my own). Is this intentional or it's a bug?

### DIFF
--- a/soem/ethercatconfig.c
+++ b/soem/ethercatconfig.c
@@ -598,6 +598,7 @@ int ecx_config_init(ecx_contextt *context, uint8 usetable)
                ECT_REG_ALCTL,
                htoes(EC_STATE_PRE_OP | EC_STATE_ACK),
                EC_TIMEOUTRET3); /* set preop status */
+            (void)ecx_statecheck(context, slave, EC_STATE_PRE_OP, EC_TIMEOUTSTATE);
          }
       }
    }
@@ -1305,6 +1306,7 @@ int ecx_config_map_group(ecx_contextt *context, void *pIOmap, uint8 group)
                   ECT_REG_ALCTL,
                   htoes(EC_STATE_SAFE_OP),
                   EC_TIMEOUTRET3); /* set safeop status */
+               (void)ecx_statecheck(context, slave, EC_STATE_SAFE_OP, EC_TIMEOUTSTATE);
             }
             if (context->slavelist[slave].blockLRW)
             {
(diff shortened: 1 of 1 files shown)


nakarlsson:
Yes, it might very well happen,  but it is better to leave the checks to the application, even with the added checks the application would need to perform the checks case one or more slaves didn't enter the expected state.

We don't build everything; it is on the TODO to reorganize the samples. https://github.com/OpenEtherCATsociety/SOEM/pull/486 is a good step in the right direction.

alexkalmuk:
@nakarlsson Thank you for this explanation! I see a little difference between controlling states from only the application and handling implicit transitions between states that happen in ethercatconfig.c. In the first case, the user requests state transition himself and it's his responsibility to handle this transition. In the second case, the user has to know for sure in which state the slave will be after `ec_config_init` (PRE_OP) or `ec_config_map` (SAFE_OP). Probably, it would be useful to mention these states in the description of the corresponding functions. Now I see that the user has to be aware of this implicit behavior if he wants to perform any transitions between slaves' states.

Should I fix `firm_update` correspondingly and create new PR? Or should I leave it as is due to reorganizing the samples?

nakarlsson:
There could be a huge performance penalty to check the states inside ec_config_init/ec_config_map compared to perform a single BRD of the states when the mentioned tasks are completed.

explicit/implicit state change are there as an option where implicate is the legacy behavior. Users have requested to have the option block that.

alexkalmuk:
It sounds reasonable, master can perform some work during transitions between states is in progress. Thanks.

alexkalmuk:
I made another fix instead of this one - https://github.com/OpenEtherCATsociety/SOEM/pull/498

Issue #497, opened 2021-04-06
DC master SysTime incorrect calculation


alexkalmuk:
https://github.com/OpenEtherCATsociety/SOEM/blob/25e36e44f37e396f359f748978878bcc4c6bc90b/soem/ethercatdc.c#L271-L272

But in osal/linux osal_current_time is implemented using CLOCK_MONOTONIC (it calls osal_gettimeofday which uses monotonic clock):

https://github.com/OpenEtherCATsociety/SOEM/blob/25e36e44f37e396f359f748978878bcc4c6bc90b/osal/linux/osal.c#L30-L36

In Linux https://linux.die.net/man/3/clock_gettime (btw, not only for Linux but for POSIX):

```
CLOCK_MONOTONIC
    Clock that cannot be set and represents monotonic time since some unspecified starting point.
```

It means ethercatdc.c cannot rely on Epoch start with current osal_current_time implementation.

nakarlsson:
that's true, an improvement would be to do as for win32. MONOTONIC is important and a must for the timers

alexkalmuk:
Do you mean this part?

https://github.com/OpenEtherCATsociety/SOEM/blob/25e36e44f37e396f359f748978878bcc4c6bc90b/osal/win32/osal.c#L37-L47

But the difference with Linux here that `GetSystemTimeAsFileTime` returns the time from Windows epoch, while MONOTONIC on Linux returns time from an unspecified moment in the past (e.g. system power on). By the way, after reading this description (https://docs.microsoft.com/en-us/windows/win32/sysinfo/system-time, the last paragraph) I am in doubts that `GetSystemTimeAsFileTime` returns monotonic clocks.

If we want to have a single function on Linux which maintains a monotonic clock and returns nanoseconds from 1 Jan, 2000, then probably we can use `gettimeofday` only once, store the result, and then add this result to MONOTONIC clocks.

nakarlsson:
I was refering to win32 having a ”int osal_getrelativetime” that would use the MONOTONIC clock on Linux , that is used for the timers. 

Get current_time/timeofday can be clock  REALTIME that doesn’t need to be strictly increasing.

alexkalmuk:
Ah, I see.

PR #499, opened 2021-04-06
Fix linux osal timers and osal_current_time

alexkalmuk:
Fixes https://github.com/OpenEtherCATsociety/SOEM/issues/497

### DIFF
--- a/osal/linux/osal.c
+++ b/osal/linux/osal.c
@@ -21,30 +21,15 @@ int osal_usleep (uint32 usec)
    return nanosleep(&ts, NULL);
 }
 
-int osal_gettimeofday(struct timeval *tv, struct timezone *tz)
-{
-   struct timespec ts;
-   int return_value;
-   (void)tz;       /* Not used */
-
-   /* Use clock_gettime to prevent possible live-lock.
-    * Gettimeofday uses CLOCK_REALTIME that can get NTP timeadjust.
-    * If this function preempts timeadjust and it uses vpage it live-locks.
-    * Also when using XENOMAI, only clock_gettime is RT safe */
-   return_value = clock_gettime(CLOCK_MONOTONIC, &ts);
-   tv->tv_sec = ts.tv_sec;
-   tv->tv_usec = ts.tv_nsec / 1000;
-   return return_value;
-}
-
 ec_timet osal_current_time(void)
 {
-   struct timeval current_time;
+   struct timespec current_time;
    ec_timet return_value;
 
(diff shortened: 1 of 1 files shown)


nakarlsson:
Given the comments, why don't you use  clock_gettime(CLOCK_REALTIME , &ts); in osal_gettimeofday ?
Otherwise it make sense. What do you say @ArthurKetels?

We could make osal_gettimeofday/osal_getrelativetime "static"

nakarlsson:
squash the commits, multiple commits are not relevant here

alexkalmuk:
> squash the commits, multiple commits are not relevant here

Yes, of course. I just think it's a good practice to squash commits only after review to monitor new fixes.

alexkalmuk:
squashed

ArthurKetels:
@nakarlsson Yes I would prefer using clock_gettime(CLOCK_REALTIME , &ts); over gettimeofday(tv, tz);. It makes our intend more clear. We would not like to depend on internal implementation of gettimeofday() because it could change or have different semantics on different OS.

nakarlsson:
@ArthurKetels and OK to re-introduce the real-time for DC time and Mbox communication? 
I believe the introduction of MONOTONIC time changed that, previously it was all gettimeofday.

ArthurKetels:
As @alexkalmuk mentioned, gettimeofday is obsolete. Using REALTIME is the preferred option. The use cases for osal_current_time() are strictly for logging and start reference of DC. 

For DC we have no other option, we want to be as close as possible to wall clock time. We know we drift from that time after a while because we sync on slave reference time, but that is a whole other matter. The application can track the drift between realtime and DC time if it needs to. Another question is if we want time-zones? I would like to have machines running in UTC only, but customers might see that differently.

alexkalmuk:
> Another question is if we want time-zones? I would like to have machines running in UTC only, but customers might see that differently.

BTW if timezones are required then `clock_gettime` probably can be used together with `localtime`/`localtime_r`.

alexkalmuk:
Hm, as @nakarlsson mentioned about a potential change in DC behavior due to moving from MONOTONIC to REALTIME clock, shouldn't slave's System Time offset be changed each time in case of master's local time (REALTIME) was adjusted?

nakarlsson:
I think the scope of this issue is covered by the current commit. @alexkalmuk do you agree? I'll merge the PR if so,

alexkalmuk:
@nakarlsson Yes, of course. Thanks for your suggestions here and in the issue, they were very useful!

PR #505, opened 2021-04-28
Report FoE error code to error list

### LINKED ISSUE #78: FoE error from the slave is not propagated
Propegate the FoE errors if one occur

yschmid-levi:
Resolves #78. 

Since I don't have an EtherCAT slave that supports FoE read, that is untested.

### DIFF
--- a/soem/ethercatfoe.c
+++ b/soem/ethercatfoe.c
@@ -45,6 +45,8 @@ typedef struct PACKED
 } ec_FOEt;
 PACKED_END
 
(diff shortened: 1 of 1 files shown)


yschmid-levi:
`context->ecaterror` is set to `TRUE` in `ecx_pusherror()` so I think we don't need to set it before the call to `ecx_pusherror()`.

https://github.com/OpenEtherCATsociety/SOEM/blob/b13f2599c1213e751c82c1df860255334839a8ec/soem/ethercatmain.c#L153-L171

PR #533, opened 2021-07-13
matching call to timeEndPeriod and timeBeginPeriod

### LINKED ISSUE #532: calls to timeBeginPeriod and timeEndPeriod do not match (win32)
According to the [Microsoft's documentation](https://docs.microsoft.com/en-us/windows/win32/api/timeapi/nf-timeapi-timebeginperiod), calls to timeBeginPeriod and timeEndPeriod must have matching arguments:

> You must match each call to timeBeginPeriod with a call to timeEndPeriod, specifying the same minimum resolution in both calls.

But they don't:

https://github.com/OpenEtherCATsociety/SOEM/blob/219670218b78acf4e4a96da70d36eeb9fc7f96ca/osal/win32/osal.c#L20

https://github.com/OpenEtherCATsociety/SOEM/blob/219670218b78acf4e4a96da70d36eeb9fc7f96ca/oshw/win32/nicdrv.c#L164

yschmid-levi:
fixes  #532

### DIFF
--- a/oshw/win32/nicdrv.c
+++ b/oshw/win32/nicdrv.c
@@ -161,7 +161,7 @@ int ecx_setupnic(ecx_portt *port, const char *ifname, int secondary)
  */
 int ecx_closenic(ecx_portt *port)
 {
-   timeEndPeriod(15);
+   timeEndPeriod(1);
 
    if (port->sockhandle != NULL)
    {


ArthurKetels:
The patch is correct but the end result does not matter. When the program closes all modifications to timePeriod are reset to default.

PR #548, opened 2021-08-31
Fix some UB cases. Closes #546

### LINKED ISSUE #546: SIGILL when compiling with `zig cc`
I wrote a small test program that uses the zig build system to compile both, the SOEM library and my test program. zig bundles clang, in my case the following version:

```
➜ zig cc --version
clang version 12.0.1 (git@github.com:ziglang/zig-bootstrap.git 8cc2870e09320a390cafe4e23624e8ed40bd363c)
Target: x86_64-unknown-linux-musl
Thread model: posix
```

The test program is very similar to the `simple_ng.c` test, and when I run it, I receive a couple of `SIGILL`s (see below). This doesn't happen if I compile the library with cmake (with gcc or clang). Apparently, zig uses different compiler flags by default (i.e. in Debug mode) and therefore found these issues.

The problems can be "resolved" with the following compiler flags.

```
-fno-delete-null-pointer-checks -fwrapv
```

However, if you agree with the findings, I can open a PR and try to fix them.

Many thanks!

--

Here, `mapb` is 31.
```
* thread #1, name = 'test1', stop reason = signal SIGILL: illegal instruction operand
    frame #0: 0x000000000023d856 test1`ecx_siigetbyte(context=0x00007ffffffdee88, slave=1, address=216) at e
thercatmain.c:383:41
   380           for(lp = 0 ; lp < cnt ; lp++)
   381           {
   382              /* set bitmap for each byte that is read */
-> 383              context->esimap[mapw] |= (1 << mapb);
```

Here, `mapb` is again 31.
```
* thread #1, name = 'test1', stop reason = signal SIGILL: illegal instruction operand
    frame #0: 0x000000000023d05c test1`ecx_siigetbyte(context=0x00007ffffffdee88, slave=2, address=319) at e
thercatmain.c:352:46
   349     {
   350        mapw = address >> 5;
   351        mapb = (uint16)(address - (mapw << 5));
-> 352        if (context->esimap[mapw] & (uint32)(1 << mapb))
```

Here, `context->slavelist[slave].inputs` is 0 (NULL).
```
* thread #1, name = 'test1', stop reason = signal SIGILL: illegal instruction operand
    frame #0: 0x00000000002280d9 test1`ecx_config_overlap_map_group(context=0x00007ffffffdee88, pIOmap=0x000
07ffffffdef40, group='\0') at ethercatconfig.c:1477:46
   1474       {
   1475          if (!group || (group == context->slavelist[slave].group))
   1476          {
-> 1477             context->slavelist[slave].inputs += context->grouplist[group].Obytes;
```

eglimi:
There are two cases of UB that are fixed in this commit.

1. In ethercatmain.c, there are two left shifts of 31:

    (1 << 31)

Because 1 is a signed int by default, the result cannot be represented
in an int. The fix is to explicitly make the 1 unsigned.

2. In ethercatconfig.c, for slaves that have no inputs, the code would
   apply an offset to a NULL pointer. The fix is to test that the slave
   has inputs available before applying the offset.

Both cases were found by clang with the help of UBSan.

### DIFF
--- a/soem/ethercatconfig.c
+++ b/soem/ethercatconfig.c
@@ -1474,7 +1474,10 @@ int ecx_config_overlap_map_group(ecx_contextt *context, void *pIOmap, uint8 grou
       {
          if (!group || (group == context->slavelist[slave].group))
          {
-            context->slavelist[slave].inputs += context->grouplist[group].Obytes;
+            if(context->slavelist[slave].Ibits > 0)
+            {
+               context->slavelist[slave].inputs += context->grouplist[group].Obytes;
+            }
          }
       }
 
--- a/soem/ethercatmain.c
+++ b/soem/ethercatmain.c
@@ -349,7 +349,7 @@ uint8 ecx_siigetbyte(ecx_contextt *context, uint16 slave, uint16 address)
    {
       mapw = address >> 5;
       mapb = (uint16)(address - (mapw << 5));
-      if (context->esimap[mapw] & (uint32)(1 << mapb))
+      if (context->esimap[mapw] & (1U << mapb))
       {
          /* byte is already in buffer */
          retval = context->esibuf[address];
@@ -380,7 +380,7 @@ uint8 ecx_siigetbyte(ecx_contextt *context, uint16 slave, uint16 address)
          for(lp = 0 ; lp < cnt ; lp++)
          {
             /* set bitmap for each byte that is read */
-            context->esimap[mapw] |= (1 << mapb);
+            context->esimap[mapw] |= (1U << mapb);
             mapb++;
             if (mapb > 31)
             {


soem/ethercatconfig.c
```diff
@@ -1474,7 +1474,9 @@ int ecx_config_overlap_map_group(ecx_contextt *context, void *pIOmap, uint8 grou
       {
          if (!group || (group == context->slavelist[slave].group))
          {
-            context->slavelist[slave].inputs += context->grouplist[group].Obytes;
+            if(context->slavelist[slave].Ibits > 0) {
```
nakarlsson, line 1477: Please use the same style for braces position as the rest of the file

soem/ethercatmain.c
```diff
@@ -349,7 +349,7 @@ uint8 ecx_siigetbyte(ecx_contextt *context, uint16 slave, uint16 address)
    {
       mapw = address >> 5;
       mapb = (uint16)(address - (mapw << 5));
-      if (context->esimap[mapw] & (uint32)(1 << mapb))
+      if (context->esimap[mapw] & (uint32)(1U << mapb))
```
hefloryd, line 352: The uint32 cast is probably not needed anymore? It is not used on line 383.

PR #552, opened 2021-09-08
Examples: show avaialable adapters if not specified

mheden:
This patch shows available adapters if not specified on the commandline for the example programs. The functionality was already available for `simple_test` but not for the others.


nakarlsson:
This is a good improvement

PR #560, opened 2021-10-20
fixed some compilation errors when compiling using clang-cl.exe

georg-emg:
When compiling the osal sub directory under Windows using clang-cl.exe, some compiler errors occurred. This patch fixes these errors, so that SOEM can be built under Windows using clang-cl.

### DIFF
--- a/osal/win32/osal.c
+++ b/osal/win32/osal.c
@@ -132,8 +132,10 @@ void osal_free(void *ptr)
 
 int osal_thread_create(void *thandle, int stacksize, void *func, void *param)
 {
-   *(OSAL_THREAD_HANDLE *)thandle = CreateThread(NULL, stacksize, func, param, 0, NULL);
-   if (!thandle)
+   HANDLE thread;
+   thread = CreateThread(NULL, (SIZE_T)stacksize, (LPTHREAD_START_ROUTINE)func, param, 0, NULL);
+   *(OSAL_THREAD_HANDLE*)thandle = thread;
+   if (thread == NULL)
    {
       return 0;
    }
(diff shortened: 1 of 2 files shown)


osal/win32/osal_win32.h
```diff
@@ -29,6 +29,7 @@
     }                                                                         \
   } while (0)
 
+struct timezone;
 int osal_gettimeofday (struct timeval *tv, struct timezone *tz);
```
nakarlsson, line 33: The declaration of osal_gettimeofday can be removed hence no need to add forward declaration of struct timezone.
I believe osal_win32.h might be removed all together

osal/win32/osal.c
```diff
@@ -130,23 +130,23 @@ void osal_free(void *ptr)
    free(ptr);
 }
 
-int osal_thread_create(void **thandle, int stacksize, void *func, void *param)
+int osal_thread_create(void *thandle, int stacksize, void *func, void *param)
 {
-   *thandle = CreateThread(NULL, stacksize, func, param, 0, NULL);
+   *(void **)thandle = CreateThread(NULL, stacksize, func, param, 0, NULL);
```
nakarlsson, line 135: I'd prefer to typecast it to PHANDLE instead of void ** , very much like the linux port


int osal_thread_create(void *thandle, int stacksize, void *func, void *param)
{
   PHANDLE phandle = thandle;
   
   *phandle = CreateThread(NULL, stacksize, func, param, 0, NULL);
   if(*phandle != NULL)
   {
      return 0;
   }
   return 1;
}

int osal_thread_create_rt(void *thandle, int stacksize, void *func, void *param)
{
   int ret;
   PHANDLE phandle = thandle;
   
   ret = osal_thread_create(phandle, stacksize, func, param);
   if (ret)
   {
      ret = SetThreadPriority(*phandle, THREAD_PRIORITY_TIME_CRITICAL);
   }
   return ret;
}

nakarlsson:
@georg-emg  , sorry for late response. Small suggestions and I don't have clang on windows so please test and give feedback.

aditmak:
@nakarlsson, suggested changes have been added and tested with clang.

PR #596, opened 2022-02-25
Some addition for cmake

KomarDL:
Add `__STDC_LIMIT_MACROS`  and `__STDC_CONSTANT_MACROS`  compile definitions for work correctly with C++ code (in case when adding library using `add_subdirectory` from C++ cmake project)

Add wpcap files to soem include list. Without it can't find <pcap.h> and other wpcap headers.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,6 +76,8 @@ add_library(soem STATIC
   ${OSHW_EXTRA_SOURCES})
 target_link_libraries(soem ${OS_LIBS})
 
+target_compile_definitions(soem PUBLIC __STDC_LIMIT_MACROS __STDC_CONSTANT_MACROS)
+
 target_include_directories(soem PUBLIC
   $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/soem>
   $<INSTALL_INTERFACE:include/soem>)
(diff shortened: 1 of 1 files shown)


nakarlsson:
ping @hefloryd

CMakeLists.txt
```diff
@@ -76,6 +78,8 @@ add_library(soem STATIC
   ${OSHW_EXTRA_SOURCES})
 target_link_libraries(soem ${OS_LIBS})
 
+target_compile_definitions(soem PUBLIC __STDC_LIMIT_MACROS __STDC_CONSTANT_MACROS)
```
hefloryd, line 81: These macros should only be set if the compiler is a C++-compiler, and preferably only if a C++-standard older than C++11 is used since it seems that this problem was fixed in C++11 (if I understand correctly). See e.g. https://sourceware.org/bugzilla/show_bug.cgi?id=15366. Have you tried with compiling with C++11?

KomarDL: Hello! I'm happy to see that my pull request has been reviewed. In my case SOEM has been included as a submodule to the project on C++17 (compiler: MSVC2019). And i had this issue. Now I haven't got access to this project :( Also I found question on SO about macroses https://stackoverflow.com/a/986584. So, i think this line is neccesary to work properly with C++ compilers. I will fix you other remarks as soon as i can

PR #603, opened 2022-03-24
Lucchartier/eathercat driver

ChartierLuc:
Tests ethercat read write. Hard coded values for AMAX-4830.


nakarlsson:
Hi, what is the rationale to add this PR to SOEM?

PR #614, opened 2022-05-12
avoid 'maybe used unitialized' warnings

notofug:
gcc : -Wmaybe-uninitialized or -Wuninitialized

Enable consuming projects to build with strict settings.

I believe the fixes should be safe, let me know if adjustments are required ;-)

### DIFF
--- a/soem/ethercateoe.c
+++ b/soem/ethercateoe.c
@@ -420,6 +420,8 @@ int ecx_EOErecv(ecx_contextt *context, uint16 slave, uint8 port, int * psize, vo
    NotLast = TRUE;
    buffersize = *psize;
    rxfragmentno = 0;
+   rxframeno = 0xff;
+   rxframeoffset = 0;
    
    /* Hang for a while if nothing is in */
    wkc = ecx_mbxreceive(context, slave, (ec_mbxbuft *)&MbxIn, timeout);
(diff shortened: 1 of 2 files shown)


nakarlsson:
I see the point in changing but would like to reproduce it, I tested gcc-11 without luck. Any tip?

notofug:
Yes sorry ; New gcc is smart enough to realize the variables will be ok and does not complain.
We use some older cross-compiling toolchains, however - and I am not sure where the breakpoints are unfortunately, but :
- gcc 7.5.0 for Armv8 (32bit) by Linaro gives warning for ethercateoe.c
- gcc 4.5.2 for PPC by Freescale gives warning for both of the files in this PR

soem/ethercateoe.c
```diff
@@ -410,9 +410,9 @@ int ecx_EOErecv(ecx_contextt *context, uint16 slave, uint8 port, int * psize, vo
    ec_EOEt *aEOEp;
    ec_mbxbuft MbxIn;
    uint16 frameinfo1, frameinfo2;
-   uint8 rxfragmentno, rxframeno;
+   uint8 rxfragmentno, rxframeno = 0xff;
```
nakarlsson, line 413: Move init to line 423 , after rxfragmentno.

soem/ethercateoe.c
```diff
@@ -410,9 +410,9 @@ int ecx_EOErecv(ecx_contextt *context, uint16 slave, uint8 port, int * psize, vo
...
    boolean NotLast;
-   int wkc, buffersize, rxframesize, rxframeoffset, eoedatasize;
+   int wkc, buffersize, rxframesize, rxframeoffset = -1, eoedatasize;
```
nakarlsson, line 415: Move init to line 423 , after rxfragmentno.

nakarlsson:
Minor cosmetics. I reproduced it with an online compiler that included 4.4.x something

notofug:
Pls check if OK now

nakarlsson:
Looks good, no warnings left on your side?

notofug:
All fine here too ;-)

nakarlsson:
thanks

PR #622, opened 2022-06-28
also configure the last SM

### LINKED ISSUE #617: Sync Manager 7 does not get configurated
Hello,
I tried to use a slave which uses Sync Manager 7 for inputs.
Unfortunately no FMMU seems to be assigned to this Sync Manager.

I tracked this down to the following lines in ecx_config_create_input_mappings:

```/* search for SM that contribute to the input mapping */
while ((SMc < (EC_MAXSM - 1)) && (FMMUdone < ((context->slavelist[slave].Ibits + 7) / 8)))
{
     ...
     /* do configuration */
     ...
}
 ```
What is the reasoning for the `(EC_MAXSM - 1)`?
When I change this to just `EC_MAXSM` a free FMMU is correctly assigned to the Sync Manger.

lbckmnn:
fixes #617

unfortunately i don't have access to a slave which actually uses SM7, so i was not able to test this specific commit/PR. 
However, i did exactly the same when i had access to a slave which used SM7 and it did work (see https://github.com/OpenEtherCATsociety/SOEM/issues/617)

### DIFF
--- a/soem/ethercatconfig.c
+++ b/soem/ethercatconfig.c
@@ -919,7 +919,7 @@ static void ecx_config_create_input_mappings(ecx_contextt *context, void *pIOmap
       }
    }
    /* search for SM that contribute to the input mapping */
-   while ((SMc < (EC_MAXSM - 1)) && (FMMUdone < ((context->slavelist[slave].Ibits + 7) / 8)))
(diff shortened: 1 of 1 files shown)


nakarlsson:
If you say it works it should be fine, Arthur suggested the fix and I agree.

Issue #626, opened 2022-07-14
possible bug in ec_readstate and ec_statecheck


lbckmnn:
I was curios how SOEM determines the states of the slaves and read the implementation of ec_readstate and ec_statecheck.
I might have found a possible bug there.
When either of these function is used with slavenumber 0 (= whole network) the AL Status code (and therefore, the ethercat state) is read with an BRD frame. 

The BRD frame seems to apply a bitwise OR.
So if slaves have different states their state is basically ORed. 
The ethercat States are mostly collision free in this regard, but apparently not entirely:

| State   | hex/dec | bin  |
|---------|---------|------|
| INIT    | 1       | 0001 |
| PRE_OP  | 2       | 0010 |
| BOOT    | 3       | 0011 |
| SAFE_OP | 4       | 0100  |
| OP      | 8       | 1000 |

As far as i can tell, if one slave is in INIT and another in PRE_OP, reading the AL Status code with a BRD would result in 0x3 which is also the State for BOOT. Is this correct?

Code from  ecx_readstate
```C

//read state of all slaves with BRD and put this in variable "bitwisestate"
...
switch (bitwisestate)
   {
      case EC_STATE_INIT:
      case EC_STATE_PRE_OP:
      case EC_STATE_BOOT:
      case EC_STATE_SAFE_OP:
      case EC_STATE_OPERATIONAL:
         allslavessamestate = TRUE;
         context->slavelist[0].state = bitwisestate;
         break;
      default:
         allslavessamestate = FALSE;
         break;
   }
    
   if (noerrorflag && allslavessamestate && allslavespresent)
   {
      /* No slave has toggled the error flag so the alstatuscode
       * (even if different from 0) should be ignored and
       * the slaves have reached the same state so the internal state
       * can be updated without sending any datagram. */
      for (slave = 1; slave <= *(context->slavecount); slave++)
      {
         context->slavelist[slave].ALstatuscode = 0x0000;
         context->slavelist[slave].state = bitwisestate;
      }
      lowest = bitwisestate;
   }

```
Wouldn't this code set the state of all slaves to BOOT if one slave is in Init and one in PRE_OP?
A similiar problem exists in ec_statecheck.

nakarlsson:
You’re right. BOOT should be removed from the case in ecx_readstate(ecx_contextt *context) for slave == 0.


[...]

PR #643, opened 2022-09-17
remove boot state from ecx_readstate and add some documentation

lbckmnn:
This addresses the issues from #626

### DIFF
--- a/soem/ethercatmain.c
+++ b/soem/ethercatmain.c
@@ -754,7 +754,7 @@ int ecx_readstate(ecx_contextt *context)
 
    switch (bitwisestate)
    {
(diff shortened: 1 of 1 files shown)


soem/ethercatmain.c
```diff
@@ -753,9 +754,9 @@ int ecx_readstate(ecx_contextt *context)
 
    switch (bitwisestate)
    {
+       //Note: BOOT State collides with PRE_OP | INIT and cannot be used here
```
nakarlsson, line 757: please use c-style comment

lbckmnn: done

PR #651, opened 2022-10-17
Corrections to create relocatable CMake package (for linux and windows)

kutij:
Now by calling CMake install, a relocatable CMake package can be created, that
- can contain Release/Debug/MinSizeRel/RelWithDebInfo versions as well, and will be linked appropriately
- can be used via simple find_package (will link the system libraries automatically, copy the additional libraries from the SOEM e.g. WinPCap these libraries are copied into the package as well)
- there is an example CMake app (only provides the slaveinfo) that gives an example to automatic clone/build/install/link of SOEM but also can be performed by installing it on your own and calling "find_package(soem)" "target_link_libraries(projectname soem)"...

The modifications are perfomed for the case of Windows and Linux. Tested on Windows 10 and Ubuntu 22.04.


kutij:
I'm not really familiar with squashing commits - I've done it in a new branch and sent a new pull request. Check it if you like it.

PR #669, opened 2022-11-29
check for invalid mailbox header length to avoid access violation

andre-comet:
When working with an Ethercat CoE slave that would occasionally send an invalid packet with mailbox size 0 during segmented transfer, SOEM causes a hard fault (access violation).

While the root cause is a bug in the slave, SOEM should handle it and signal an error instead of crashing due to an access violation.

The issue is caused in line 244 in the calculation of Framedatasize:
`Framedatasize = etohs(aSDOp->MbxHeader.length) - 3`
If `MbxHeader.length` is zero, the UINT16 value wraps to 65533 (0 - 3 = 65535 - 2). Faulty `Framedatasize` of 65533 is then passed as size argument to memcpy, causing the access violation

Invalid packet:
image

Framedatasize is 65533 when passed as size parameter to memcpy in line 257 of ethercatcoe.c:
image

My change simply checks if the `-3` calculation would cause a wrap of the UINT. If it would, `ecx_packeterror()` is called.

I tested this change with the pysoem wrapper on windows and it now throws a packet error exception which can be handled, instead of causing an access violation which crashed the Ethercat master entirely.


ArthurKetels (changes requested):
Thanks for catching this bug. The solution is however not enough to fix all cases. 
Example 1: MbxHeader.Length  - 3 > buffer size
Example 2: More segmented transfers than buffer size. Each segment being perfectly valid.
The solution should be that each memcpy in the segmented transfer is checked for bounds before executing.

PR #750, opened 2023-10-13
Remove variable for unused parameter.

rtmagj:
ecatcheck doesn't use the parameter sent to it. 
On system with no ctime, ctime caused undeclared identifier when building.
Replacing ctime with NULL.


hefloryd (approved):
Looks good to me. Maybe remove the commented call to pthread_create instead.

PR #769, opened 2023-12-06
Allow FoE write of 0byte file

### LINKED ISSUE #766: ec_FOEwrite: Allow sending a 0byte file to "clear" a slave file in file system, if file is present.
Hi,

I got scenario where a customer try to "clear" a file in the slave filesystem if it exists by sending a 0 size file with given filename. 
Doing this will make the SOES FoE state machine end up in an error state since no DATA packet is sent.
Seems SOEM ecx_FOEwrite doesn't support that out-of-the-box since it will not send the 0 size DATA packet due to the file line below. My suggestion was to add a SDO that would do this in a controlled way.

https://github.com/OpenEtherCATsociety/SOEM/blob/c5db2741ee131decd80c94b51afda5f947ce7f95/soem/ethercatfoe.c#L217

However, I can't find anything in the EtherCAT specification that would disallow sending a 0 byte file, To solve this one would initialize dofinalzero to TRUE and it would work as expected. 

Any thoughts?

nakarlsson:
If FoE write of a file is done, no check
of incoming size is performed. If the
file is 0byte, a FoE Write Request is
sent but no ACK since dofinalzero
is initialized to FALSE.

fixes #766

### DIFF
--- a/soem/ethercatfoe.c
+++ b/soem/ethercatfoe.c
@@ -214,7 +214,7 @@ int ecx_FOEwrite(ecx_contextt *context, uint16 slave, char *filename, uint32 pas
    ec_clearmbx(&MbxOut);
    aFOEp = (ec_FOEt *)&MbxIn;
    FOEp = (ec_FOEt *)&MbxOut;
-   dofinalzero = FALSE;
+   dofinalzero = TRUE;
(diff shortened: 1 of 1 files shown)


ArthurKetels (approved):
Good solution, simple.

PR #782, opened 2024-01-23
tests: fix linked list leaks

JohannesKauffmann:
Don't overwrite the original pointer returned by ec_find_adapters(), otherwise the linked list is leaked. Instead, save original pointer to the head of the list use and `adapters` as a temporary variable. Pass the original pointer to ec_free_adapters().

For win32, ec_free_adapters() was missing entirely.

### DIFF
--- a/test/linux/eepromtool/eepromtool.c
+++ b/test/linux/eepromtool/eepromtool.c
@@ -457,6 +457,7 @@ int main(int argc, char *argv[])
    else
    {
       ec_adaptert * adapter = NULL;
+      ec_adaptert * head = NULL;
 
       printf("Usage: eepromtool ifname slave OPTION fname|alias\n");
       printf("ifname = eth0 for example\n");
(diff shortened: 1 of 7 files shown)


JohannesKauffmann:
ping @ArthurKetels @nakarlsson

JohannesKauffmann:
@hefloryd

test/linux/eepromtool/eepromtool.c
```diff
@@ -456,7 +456,7 @@ int main(int argc, char *argv[])
    }
    else
    {
-      ec_adaptert * adapter = NULL;
+      ec_adaptert * adapter, head = NULL;
```
hefloryd, line 459: This doesn't compile, because head is not a pointer to ec_adaptert. Prefer not to declare more than one variable per line to avoid this type of mistake.

JohannesKauffmann: Thank you for the review. How embarrassing, I was sure I'd tested this beforehand. I've updated the PR to declare `head` on a separate line.

PR #815, opened 2024-05-29
fix: add missing parameters to printf

HenrikCohrs:


test/linux/aliastool.c
```diff
@@ -368,7 +368,7 @@ void eepromtool(char *ifname, int slave, int mode, char *fname)
             if (mode == MODE_WRITEALIAS)
 			{
 			  if(eeprom_writealias(slave, alias))
-			    printf("Alias %4.4X written successfully to slave %d\n");
+			    printf("Alias %4.4X written successfully to slave %d\n", (unsigned)alias, slave);
```
nakarlsson, line 371: Is the explicit cast necessary? There are other similar calls for fprintf not having it.

HenrikCohrs: No, the explicit cast is not necessary.  
I checked the documentation, and since the argument undergoes [integer promotion](https://en.cppreference.com/w/c/language/conversion#Default_argument_promotions), it is not needed.

HenrikCohrs:
I signed the CLA, but something does not seem to be working. In GitHub, it still says, that I need to sign it.

Screenshot 2025-10-01 212848

PR #831, opened 2024-07-01
fix windows warning

notofug:
avoid : error C4703: potentially uninitialized local pointer variable 'prev_adapter' used


hefloryd:
Merged, thanks

PR #835, opened 2024-07-11
Rename PACKED macro to SOEM_PACKED to avoid collision with fmtlib

rtzoeller:
There is an unfortunate collision between SOEM and [fmtlib](https://github.com/fmtlib/fmt), where SOEM's definition of a `PACKED` macro collides with a [template parameter in fmtlib](https://github.com/fmtlib/fmt/blob/92cdbbae06fe43f87a6c59eb1d623bb5abc8405d/include/fmt/base.h#L1649).

This causes some rather inscrutable compile errors, and isn't handled by the existing `#ifndef` check as fmtlib's usage is not a macro.

### DIFF
--- a/doc/tutorial.txt
+++ b/doc/tutorial.txt
@@ -326,7 +326,7 @@ set_output_int16(EL4001_1,0,slave_EL4001_1.out1);
 Target can handle non aligned pointers to the IOmap
 \code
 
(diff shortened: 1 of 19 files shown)


hefloryd:
This was fixed in ed3ce75, thanks.

PR #837, opened 2024-07-17
Fix missing return value checks in `ecx_setupnic()` for linux

hmfhmf:
Without this fix, it has been observed that `ecx_setupnic` in some cases erroneously returns success when trying to setup a NIC that doesn't exist.


hefloryd:
Merged, thanks

PR #843, opened 2024-07-31
CMakeLists.txt: Add possibility to build library as shared

DasRoteSkelett:
This adds the option to build the soem library also as shared object. Note that the default is untouched, so by default, it will still build a static library.

The [BUILD_SHARED_LIBS](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html) is the standard way of doing it, many tools use this semantics to influence the output to be the desired type (whichever you want).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,8 @@ if (CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
   set(CMAKE_INSTALL_PREFIX ${CMAKE_CURRENT_LIST_DIR}/install)
 endif()
 
+option(BUILD_SHARED_LIBS "Build using shared libraries" OFF)
+
 set(SOEM_INCLUDE_INSTALL_DIR include/soem)
 set(SOEM_LIB_INSTALL_DIR lib)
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -75,11 +77,17 @@ file(GLOB SOEM_HEADERS soem/*.h)
...
 
-add_library(soem STATIC
+add_library(soem
   ${SOEM_SOURCES}
   ${OSAL_SOURCES}
   ${OSHW_SOURCES}
   ${OSHW_EXTRA_SOURCES})
+
+if (BUILD_SHARED_LIBS)
+  set_target_properties(soem PROPERTIES VERSION ${PROJECT_VERSION})
```
hefloryd, line 87: Should this be:

    set_target_properties(mylib PROPERTIES VERSION ${PROJECT_VERSION} SOVERSION ${PROJECT_VERSION_MAJOR})

? See https://cmake.org/cmake/help/v3.29/prop_tgt/VERSION.html#prop_tgt:VERSION

DasRoteSkelett: Thanks, good catch, fixed it.

hefloryd:
Merged, thanks

PR #891, opened 2025-02-27
Add AL status codes from ETG.1020

rtzoeller:
Add the additional AL status codes defined by ETG.1020 to `ec_ALstatuscodelist`.


rtzoeller:
ETG.1020 also shows `0x002E` as "Cycle time too small", rather than "Invalid input FMMU configuration". I am happy to make this change as well if desired.

Issue #374, opened 2020-03-03
Limit expected working counter if the slave device have multiple sync managers


nakarlsson:
Limit the expected working counter if the slave device have multiple sync managers. SOEM should not add +1 to the expected working counter for each used inputs/ouputs sync manager.

Ref: maximum working counter 3 per ESC for RW commands. 
https://www.ethercat.org/memberarea/forum/forum_posts.asp?TID=586&KW=multiple+sync+manager&title=working-counter-increment

nakarlsson:
fixed in #375

PR #907, opened 2025-04-14
fix single device inputs/outputs overruning segment size

ddi-kkugler:
certain modular devices can have process data that exceeds a segment size. Add code to allow multiple segments within a single device.

### DIFF
--- a/soem/ethercatconfig.c
+++ b/soem/ethercatconfig.c
@@ -898,7 +898,6 @@ static void ecx_config_create_input_mappings(ecx_contextt *context, void *pIOmap
 {
    int BitCount = 0;
    int FMMUdone = 0;
-   int AddToInputsWKC = 0;
    uint16 ByteCount = 0;
    uint16 FMMUsize = 0;
    uint8 SMc = 0;
@@ -1003,9 +1002,6 @@ static void ecx_config_create_input_mappings(ecx_contextt *context, void *pIOmap
          /* program FMMU for input */
          ecx_FPWR(context->port, configadr, ECT_REG_FMMU0 + (sizeof(ec_fmmut) * FMMUc),
             sizeof(ec_fmmut), &(context->slavelist[slave].FMMU[FMMUc]), EC_TIMEOUTRET3);
-         /* Set flag to add one for an input FMMU,
-            a single ESC can only contribute once */
-         AddToInputsWKC = 1;
       }
       if (!context->slavelist[slave].inputs)
       {
@@ -1031,10 +1027,6 @@ static void ecx_config_create_input_mappings(ecx_contextt *context, void *pIOmap
       FMMUc++;
    }
    context->slavelist[slave].FMMUunused = FMMUc;
-
-   /* Add one WKC for each input segment if flag is true */
-   if (AddToInputsWKC)
-      context->grouplist[group].inputsWKC += 1 + (FMMUsize / (EC_MAXLRWDATA - EC_FIRSTDCDATAGRAM));
 }
 
 static void ecx_config_create_output_mappings(ecx_contextt *context, void *pIOmap,
@@ -1042,7 +1034,6 @@ static void ecx_config_create_output_mappings(ecx_contextt *context, void *pIOma
 {
    int BitCount = 0;
    int FMMUdone = 0;
-   int AddToOutputsWKC = 0;
    uint16 ByteCount = 0;
    uint16 FMMUsize = 0;
    uint8 SMc = 0;
@@ -1141,9 +1132,6 @@ static void ecx_config_create_output_mappings(ecx_contextt *context, void *pIOma
          /* program FMMU for output */
          ecx_FPWR(context->port, configadr, ECT_REG_FMMU0 + (sizeof(ec_fmmut) * FMMUc),
             sizeof(ec_fmmut), &(context->slavelist[slave].FMMU[FMMUc]), EC_TIMEOUTRET3);
-         /* Set flag to add one for an output FMMU,
-            a single ESC can only contribute once */
-         AddToOutputsWKC = 1;
       }
       if (!context->slavelist[slave].outputs)
       {
@@ -1170,9 +1158,6 @@ static void ecx_config_create_output_mappings(ecx_contextt *context, void *pIOma
       FMMUc++;
    }
    context->slavelist[slave].FMMUunused = FMMUc;
-   /* Add one WKC for each output segment if flag is true */
-   if (AddToOutputsWKC)
-      context->grouplist[group].outputsWKC += 1 + (FMMUsize / (EC_MAXLRWDATA - EC_FIRSTDCDATAGRAM));
 }
 
 static int ecx_main_config_map_group(ecx_contextt *context, void *pIOmap, uint8 group, boolean forceByteAlignment)
@@ -1184,6 +1169,7 @@ static int ecx_main_config_map_group(ecx_contextt *context, void *pIOmap, uint8 
    uint32 diff;
    uint16 currentsegment = 0;
    uint32 segmentsize = 0;
+   uint32 segmentmaxsize = (EC_MAXLRWDATA - EC_FIRSTDCDATAGRAM); /* first segment must account for DC overhead */
 
    if ((*(context->slavecount) > 0) && (group < context->maxgroup))
    {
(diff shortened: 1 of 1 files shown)


ddi-kkugler:
I sent in my contributors agreement on April 11th, what else do I need to do to have this reviewed?

nakarlsson:
Can you provide the config to trigger this? A slaveinfo output would do

ddi-kkugler:
Output from same device with output larger than one segment. Device has modular profile that can be quite large.
```
SOEM (Simple Open EtherCAT Master)
Slaveinfo
Starting slaveinfo
ec_init on enx3c8cf8600c1c succeeded.
1 slaves found and configured.
Calculated workcounter 3

Slave:1
 Name:IO Controllers
 Output size: 5856bits
 Input size: 12216bits
 State: 4
 Delay: 0[ns]
 Has DC: 1
 DCParentport:0
 Activeports:0.1.0.0
 Configured address: 1001
 Man: 0000076b ID: 00000064 Rev: 00041180
 SM0 A:1000 L: 128 F:00010026 Type:1
 SM1 A:1600 L: 128 F:00010022 Type:2
 SM2 A:1c00 L: 732 F:00010064 Type:3
 SM3 A:5600 L:1527 F:00010020 Type:4
 FMMU0 Ls:00000000 Ll: 732 Lsb:0 Leb:7 Ps:1c00 Psb:0 Ty:02 Act:01
 FMMU1 Ls:000002dc Ll:1527 Lsb:0 Leb:7 Ps:5600 Psb:0 Ty:01 Act:01
 FMMUfunc 0:1 1:2 2:3 3:0
 MBX length wr: 128 rd: 128 MBX protocols : 0c
 CoE details: 2f FoE details: 01 EoE details: 00 SoE details: 00
 Ebus current: 0[mA]
 only LRD/LWR:0
End slaveinfo, close socket
End program
```

ddi-kkugler:
I just noticed that with this large config the actual wkc returned is 4 instead of 3. The expected wkc is still calculated as 3. Looking through spec docs now to see if there is any mention about what should happen in this case.

ddi-kkugler:
> I just noticed that with this large config the actual wkc returned is 4 instead of 3. The expected wkc is still calculated as 3. Looking through spec docs now to see if there is any mention about what should happen in this case.

ETG1000.4 indicates that the wkc is associated with each ECAT PDU. Since adding additional segments is adding more PDUs for the process data the WKC should be updated accordingly. For the slave config I have posted above the correct WKC should be 4.

nakarlsson:
Have you checked/verified the overlapped API also?

ddi-kkugler:
I don't know enough to know how to test that. If I try calling config_overlap_map then the device I have seems to not report data correctly.

nakarlsson:
> I don't know enough to know how to test that. If I try calling config_overlap_map then the device I have seems to not report data correctly.
> 
> **EDIT**: Using send_overlap_processdata with config_overlap_map the system works as expected. This is a test with only a single device on the ecat network.

Was the initial test that didn't with

ddi-kkugler:
> Sounds good, I assume first try using config_overlap_map was done using send_processdata?

Yes, exactly. Once I changed both over it worked as expected.

ddi-kkugler:
Let me know anything else you want to test/try before getting this merged.

soem/ethercatconfig.c
```diff
@@ -1170,9 +1170,9 @@ static void ecx_config_create_output_mappings(ecx_contextt *context, void *pIOma
       FMMUc++;
    }
    context->slavelist[slave].FMMUunused = FMMUc;
-   /* Add one WKC for an output if flag is true */
+   /* Add one WKC for each output segment if flag is true */
    if (AddToOutputsWKC)
-      context->grouplist[group].outputsWKC++;
+      context->grouplist[group].outputsWKC += 1 + (FMMUsize / (EC_MAXLRWDATA - EC_FIRSTDCDATAGRAM));
```
nakarlsson, line 1175: * +2 right?
* Also, add parentheses in the right hand term  (1 + (FMMUsize / (EC_MAXLRWDATA - EC_FIRSTDCDATAGRAM)));
* Perhaps add curly braces to make it more clear

soem/ethercatconfig.c
```diff
@@ -1228,8 +1228,13 @@ static int ecx_main_config_map_group(ecx_contextt *context, void *pIOmap, uint8
                   if (currentsegment < (EC_MAXIOSEGMENTS - 1))
                   {
                      currentsegment++;
-                     segmentsize = diff;
                   }
+                  while (diff > (EC_MAXLRWDATA - EC_FIRSTDCDATAGRAM) && currentsegment < (EC_MAXIOSEGMENTS - 1))
```
nakarlsson, line 1232: I wonder if there are a clever way to split a "big" segment? I did a sample slave with 4B Out/1540B In.
The logic will create 3 frames when 2 frames easily is enough, but sometimes the clever logic comes with a cost in complexity to maintain and test.

When running overlapped the DC frame is sent alone in forst segement.

Any idea?

soem/ethercatconfig.c
ddi-kkugler: > +2 right?

All the examples take this value and multiply by 2, you can see the code before was only incrementing by one. Changing that will break the contract with users. 

> Also, add parentheses in the right hand term (1 + (FMMUsize / (EC_MAXLRWDATA - EC_FIRSTDCDATAGRAM)));
> Perhaps add curly braces to make it more clear

I can refactor the code to remove the AddToOutputWKC/AddToInputWKC altogether, which would be my preference.

soem/ethercatconfig.c
ddi-kkugler: > Also, EC_MAXLRWDATA - EC_FIRSTDCDATAGRAM , EC_FIRSTDCDATAGRAM is only relevant for first frame.

The code before was/is taking EC_FIRSTDCDATAGRAM into account on all segments, I can refactor and change that if you're confident it doesn't need this. 

> I wonder if there are a clever way to split a "big" segment? I did a sample slave with 4B Out/1540B In.
The logic will create 3 frames when 2 frames easily is enough, but sometimes the clever logic comes with a cost in complexity to maintain and test.

The code right now will pack a segment with as many slaves as will fit, it should be easy to extend that for this use case.

soem/ethercatconfig.c
nakarlsson: > > +2 right?
> 
> All the examples take this value and multiply by 2, you can see the code before was only incrementing by one. Changing that will break the contract with users.
> 
> > Also, add parentheses in the right hand term (1 + (FMMUsize / (EC_MAXLRWDATA - EC_FIRSTDCDATAGRAM)));
> > Perhaps add curly braces to make it more clear
> 
> I can refactor the code to remove the AddToOutputWKC/AddToInputWKC altogether, which would be my preference.

* My bad, +1 it is
* No problem to refactor if you got an idea, the issue addressed was described in #374

soem/ethercatconfig.c
nakarlsson: > > Also, EC_MAXLRWDATA - EC_FIRSTDCDATAGRAM , EC_FIRSTDCDATAGRAM is only relevant for first frame.
> 
> The code before was/is taking EC_FIRSTDCDATAGRAM into account on all segments, I can refactor and change that if you're confident it doesn't need this.
> 
> > I wonder if there are a clever way to split a "big" segment? I did a sample slave with 4B Out/1540B In.
> > The logic will create 3 frames when 2 frames easily is enough, but sometimes the clever logic comes with a cost in complexity to maintain and test.
> 
> The code right now will pack a segment with as many slaves as will fit, it should be easy to extend that for this use case.

We can keep EC_FIRSTDCDATAGRAM to keep it simple, how would you otherwise be able to calculate an adjustment of WKC correctly? You don't know what segment it will belong to.

The logic should avoid splitting slave data between segments when possible.

ddi-kkugler: The code is refactored so the wkc is cleared an incremented all within the config mapping functions rather than spread across separate functions as before. It will also pack large slave data into the current segment to potentially avoid using additional segments. 

> The logic should avoid splitting slave data between segments when possible.

I set it up this way and I'll agree this is just a preference for what "feels right". For what it's worth, the code can be shorter (and network traffic optimized) by always packing as tightly as possible and I can't find any spec that says a slave _must_ be handled in a single PDU. The downside being that wireshark dumps _might_ be harder to decipher with more optimized packing. 

> a cost in complexity to maintain and test

You mentioned maintenance and testing a few days ago, how do you test this code? Is there a way to use SOES and run some sort of automated testing?

nakarlsson: > The code is refactored so the wkc is cleared an incremented all within the config mapping functions rather than spread across separate functions as before. It will also pack large slave data into the current segment to potentially avoid using additional segments.
> 
> > The logic should avoid splitting slave data between segments when possible.
> 
> I set it up this way and I'll agree this is just a preference for what "feels right". For what it's worth, the code can be shorter (and network traffic optimized) by always packing as tightly as possible and I can't find any spec that says a slave _must_ be handled in a single PDU. The downside being that wireshark dumps _might_ be harder to decipher with more optimized packing.

The rationale is to make Wireshark dumps as easy to dissect as possible.

> 
> > a cost in complexity to maintain and test
> 
> You mentioned maintenance and testing yesterday, how do you test this code? Is there a way to use SOES and run some sort of automated testing?

I've created two SOES slaves on XMC48 that is 4B Out/1540B In and 1540B Out/4B In. I run manual tests.

ddi-kkugler: Thank you for the info, let me know if there is anything else you want me to do.

nakarlsson: It looks good! Nice work.
I'd like to do some more regression testing on bigger networks but I'll be out of office until next Monday, thank you for understanding it'll take a few more days before merge.

PR #950, opened 2026-01-21
Fix pointer assignment for emergency message handling

jixian79:
this will cause normal SDO response is mistakenly recognized as an Emergency message。


hefloryd:
Merged, thanks

Issue #948, opened 2025-12-25
Error "Segmentation fault" in slaveinfo


kkmserver:
MS300 Device (CMM-EC03 Card)

The ESI file contains the following line:
`<CoE SdoInfo="false" PdoConfig="true" PdoAssign="true" CompleteAccess="false" SegmentedSdo="false"/>`
So, as I understand it, the device doesn't provide information on COE objects.

When running slaveinfo, I get this error in ec_main.c, line 1675:
`wkc = ecx_FPRD(&context->port, configadr, mbxro, mbxl, mbxin, EC_TIMEOUTRET);`
Because mbxin == NULL

And mbxin was previously obtained through the function аt line 1666
`mbxin = ecx_getmbx(context); `

But nothing was received in ecx_getmbx – because The device doesn't return information about COE objects.
And slaveinfo crashes with the error "Segmentation fault."

hefloryd:
Fixed by #976

PR #976, opened 2026-06-09
fix: Fix NULL dereference in ecx_mbxreceive when mailbox pool is empty

Eksan0325:
## Summary
- Check the return value of `ecx_getmbx()` before dereferencing `mbxin`
- Set `*mbx` to NULL and return 0 when no mailbox buffer is available

## Problem
When the mailbox pool is exhausted, `ecx_mbxreceive()` dereferenced a NULL
pointer and caused a segfault in `slaveinfo` (reported in #948).

## Solution
Return 0 on failure, consistent with existing mailbox API semantics
(`wkc > 0` indicates success).

### DIFF
--- a/src/ec_main.c
+++ b/src/ec_main.c
@@ -1671,7 +1671,12 @@ int ecx_mbxreceive(ecx_contextt *context, uint16 slave, ec_mbxbuft **mbx, int ti
       if ((wkc > 0) && ((SMstat & 0x08) > 0)) /* read mailbox available ? */
       {
          mbxro = slavelist->mbx_ro;
-         mbxin = ecx_getmbx(context);
+         mbxin = ecx_getmbx(context);  
+         if (mbxin == NULL) /* imbxin=NULL will cause process to crash. fix Issue #948*/
+         { 
+             *mbx = NULL;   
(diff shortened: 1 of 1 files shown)


src/ec_main.c
```diff
@@ -1671,7 +1671,12 @@ int ecx_mbxreceive(ecx_contextt *context, uint16 slave, ec_mbxbuft **mbx, int ti
       if ((wkc > 0) && ((SMstat & 0x08) > 0)) /* read mailbox available ? */
       {
          mbxro = slavelist->mbx_ro;
-         mbxin = ecx_getmbx(context);
+         mbxin = ecx_getmbx(context);  
+         if (mbxin == NULL) /* imbxin=NULL will cause process to crash. fix Issue #948*/
```
hefloryd: Please remove this comment as it will refer to a historical problem once this code is merged.

hefloryd:
Thanks, looks good. Please update the comment as requested and perhaps also add some more info to the commit message.

src/ec_main.c
Eksan0325: OK, I added another commit in this PR

PR #978, opened 2026-06-11
fix: Fix #977: Win32 Platform OSAL thread_create and thread_priority bug

### LINKED ISSUE #977: Win32 OSAL: incorrect thread creation failure check and invalid handle passed to SetThreadPriority
osal/win32/osal.c contains two latent bugs in osal_thread_create() and osal_thread_create_rt(). They affect any Windows build that uses background threads (e.g. ec_sample, eni_test, eoe_test, and multi-threaded CoE mapping when EC_MAX_MAPT > 1).

These issues are not linked to an existing GitHub Issue — they were found during code review of the Win32 OSAL layer. (Related but distinct: #546 discussed undefined behavior with zig cc, not this Win32 thread handle logic.)

# Bug 1: CreateThread failure is never detected
Current code (osal/win32/osal.c):
```cpp
int osal_thread_create(void *thandle, int stacksize, void *func, void *param)
{
   *(OSAL_THREAD_HANDLE *)thandle = CreateThread(NULL, stacksize, func, param, 0, NULL);
   if (!thandle)   /* checks the pointer argument, not CreateThread's return value */
   {
      return 0;
   }
   return 1;
}
```
## Problem:

- thandle is the address of the caller's HANDLE variable (e.g. &thread1). It is never NULL in SOEM call sites.
- When CreateThread() fails, it returns NULL, but the function still returns 1 (success).
- Callers assume the thread was created and continue execution.

## Impact:

- Silent thread creation failure on Windows.
- Caller may proceed without a running worker thread (e.g. PDO cycle thread or ecatcheck error-recovery thread in samples).
- 
# Bug 2: SetThreadPriority receives wrong argument in RT path
Current code (osal/win32/osal.c):
```cpp
int osal_thread_create_rt(void *thandle, int stacksize, void *func, void *param)
{
   int ret;
   ret = osal_thread_create(thandle, stacksize, func, param);
   if (ret)
   {
      ret = SetThreadPriority(thandle, THREAD_PRIORITY_TIME_CRITICAL);
      /*                    ^^^^^^^ pointer to HANDLE storage, not a thread HANDLE */
   }
   return ret;
}
```
## Problem:

- SetThreadPriority() expects a HANDLE to the thread (thread1), not void *thandle (&thread1).
- Real-time priority is never applied to the created thread.
- SetThreadPriority almost always fails when given an arbitrary pointer value.

## Impact:
- osal_thread_create_rt() does not actually raise thread priority on Win32.
- Used by ec_sample for the cyclic PDO thread — RT scheduling intent is ineffective on Windows.

## Expected API behavior (per osal.h)
- Return 1 on success, 0 on failure.
- Write the created thread HANDLE into *(OSAL_THREAD_HANDLE *)thandle.
- On CreateThread failure, write NULL into the caller's handle variable and return 0.

Eksan0325:
fix(win32): detect CreateThread failure and pass correct handle to SetThreadPriority

osal_thread_create() checked thandle instead of CreateThread's return value,
so thread creation failures were reported as success.

osal_thread_create_rt() passed the thandle pointer to SetThreadPriority
instead of the thread HANDLE, so real-time priority was never applied.

Fixes #977

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ install
 /doc/html
 tags
 .vscode
+.vs
--- a/osal/win32/osal.c
+++ b/osal/win32/osal.c
@@ -132,8 +132,10 @@ void osal_free(void *ptr)
 
 int osal_thread_create(void *thandle, int stacksize, void *func, void *param)
 {
-   *(OSAL_THREAD_HANDLE *)thandle = CreateThread(NULL, stacksize, func, param, 0, NULL);
-   if (!thandle)
+   HANDLE thread;
+   thread = CreateThread(NULL, (SIZE_T)stacksize, (LPTHREAD_START_ROUTINE)func, param, 0, NULL);
+   *(OSAL_THREAD_HANDLE*)thandle = thread;
+   if (thread == NULL)
    {
       return 0;
    }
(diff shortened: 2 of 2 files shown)


hefloryd:
Thanks, looks good. Can you update the commit message with the info in the pull request?

Eksan0325:
@hefloryd  Thanks for the review, I've updated the commit message with the details.

